PR #8, opened 2015-09-08
add bower support

mwcz:
For your consideration!  I created a bower.json file so p5.play can be installed with bower.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,22 @@
+{
+  "name": "p5.play",
+  "version": "0.0.0",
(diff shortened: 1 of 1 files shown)


karneaud:
Tried to `install bower p5.play --save` and got

> ENOTFOUND Package p5.play not found

islemaster:
Thanks for reporting!  I've opened issue #61 to investigate.

mwcz:
It looks like p5.play isn't published in bower.  Until that happens, this should work for you:

    bower install --save git@github.com:molleindustria/p5.play.git

PR #13, opened 2015-10-31
Group.removeSprites() removes all sprites.

### LINKED ISSUE #7: Group.removeSprites does not remove all sprites?
The following codes remains two sprites.
Am I missing something?
```javascript
// Group
var pumpkins;

function setup() {
  createCanvas(300, 300);
  pumpkins = new Group();
  for(var i = 0 ; i < 5 ; i++){
    var sprite = createSprite(random(0, width), random(0, height / 2)); 
    sprite.velocity.y = 1; 
    pumpkins.add(sprite);
  }
}

function draw() {
  background(255);
  drawSprites();
}

function mouseClicked() {
   pumpkins.removeSprites();
}
```

The following "removeSprites" code in p5.play.js looks fine.
("@method clear" should be "@method removeSprites")
```javascript
  /**
  * Removes all the sprites in the group
  * from the scene.
  *
  * @method clear
  */
  array.removeSprites = function() {
    
    for(var i = 0; i<array.length; i++)
      array[i].remove();
    
    array.length = 0;
  }
```

toolness:
This fixes #7.

This is the ideal kind of thing I'd like to make a unit test for, but p5.play doesn't seem to have a unit test suite... But that seems like it should be done in a separate issue/PR?

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -2387,11 +2387,10 @@ function Group() {
   * @method clear
   */
   array.removeSprites = function() {
+    var arrayLength = array.length;
 
-    for(var i = 0; i<array.length; i++)
-      array[i].remove();
-
-    array.length = 0;
+    for(var i = 0; i<arrayLength; i++)
+      array[0].remove();
   }
 
   /**


lib/p5.play.js
```diff
@@ -2387,11 +2387,10 @@ function Group() {
   * @method clear
   */
   array.removeSprites = function() {
+    var arrayLength = array.length;
 
-    for(var i = 0; i<array.length; i++)
-      array[i].remove();
-
-    array.length = 0;
```
toolness: This line is unnecessary, I think, since `array.length` already gets to zero by the time we've called `remove()` on all its items.

PR #22, opened 2016-01-31
Add support for p5 instance mode

### LINKED ISSUE #12: Instance mode with p5.play
Hi, there.

I tried p5.js and p5.play in instance mode, like this.

However, it stay loading...

Please tell me, is p5.play running in Instance mode?

```html
<body>
 	<div id = "p5Canvas"></div>
 	<script src="http://cdnjs.cloudflare.com/ajax/libs/p5.js/0.4.7/p5.min.js"></script>
 	<script src="/js/p5.play.js"></script>
 	<script>
 	var img
 	var s = function(sketch) {
 		var x = 100; 
 		var y = 100;
 		sketch.setup = function() {
 			sketch.createCanvas(800,400);
 			sketch.background(200);
 			sketch.fill(255);
 			img = sketch.createSprite(400, 200, 50, 50);
 		};
 
 		sketch.draw = function() {
 			sketch.drawSprites(img);
 		};
 	};
 
 	var myp5 = new p5(s, "p5Canvas");
 	</script>
 </body>
```

toolness:
This adds support for p5 instance mode in p5.play, which fixes #12.

## Global Sketch Properties

A number of variables, like `allSprites`, were formerly global, but now need to be bound to the lifetime of a particular sketch instance. I found doing this a bit tricky because there seems to be no easy way to initialize such properties upon sketch initialization; I've documented this limitation and my workaround for it at https://github.com/processing/p5.js/issues/1263.

## Constructor Changes

One particular challenge of this feature is the fact that almost all of the p5.play classes, with the notable exception of `Group`, now need to be bound to a `p5` instance; I've changed their constructors to take this instance as their first argument. However, this makes it a bit tricky to support legacy code.

~~Ideally, we should simply expose the instantiation of these classes via factory functions on `p5.prototype`, which is already being done with `createSprite`, `loadAnimation`, and `loadSpriteSheet`--however, the constructors of all these classes are also [publicly documented](http://p5play.molleindustria.org/docs/index.html), which means that it's entirely possible for legacy code to be instantiating them directly--for example, the `examples/sprites_with_sheet.js` example directly calls `new SpriteSheet()` at one point instead of using `loadSpriteSheet()`.~~

~~An alternative here is to simply *not* support legacy code when it's problematic, and increment the major version number in the next release as per [semantic versioning](http://semver.org/) to indicate backwards-incompatible API changes.~~

As per @islemaster's [suggestion](https://github.com/molleindustria/p5.play/pull/22#issuecomment-185255472), we're exposing curried versions of the constructors on all created p5 instances, so legacy code still works!

## Formerly Global Variables

In order for this to work properly, I had to enclose the entire library in an [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression). Previously, only about half of the library was enclosed in one, which meant that some of its private implementation functions, like `construct`, were actually in global scope (see below for a full list). While all the examples in the `examples/` directory still work fine, I hope that third-party code still works. Might be a good idea to QA this somehow.

The following were formerly globals but are now enclosed in an IIFE:

  * `Sprite` (now exposed via `p5.prototype.Sprite`)
  * `Camera` (now exposed via `p5.prototype.Camera`)
  * `cameraPush`
  * `cameraPop`
  * `Group` (now exposed via `p5.prototype.Group`)
  * `CircleCollider`
  * `AABB`
  * `Animation` (now exposed via `p5.prototype.Animation`)
  * `SpriteSheet` (now exposed via `p5.prototype.SpriteSheet`)
  * `Quadtree`
  * `updateTree`
  * `construct`

Note that even for the items that are now exposed via `p5.prototype`, they won't become globals until and unless p5 enters global mode on page load.

## Examples Changes

Nothing was changed in the examples, because I wanted to make sure p5.play still works with legacy code. ~~However, I *did* get rid of the extra copy of `p5.play.js` in f1363d146de27dd8a5f7f2b91fa1c2525788ff65 so that we don't need to pollute the repository's commit history by constantly updating two separate copies of the library.~~ This is still true, but has been spun out into #36.

## Other Notes

Of great help here was:

```
jshint lib/p5.play.js | grep "is not defined"
```

This logs any undefined globals, which was useful in finding p5 global methods/properties that needed to be bound to p5 instances.

## To Do

- [X] Add more tests.
- [x] Consider exposing [`Animation`](http://p5play.molleindustria.org/docs/classes/Animation.html), at least in a deprecated way, as its constructor is currently publicly documented.
- [x] Consider exposing [`Camera`](http://p5play.molleindustria.org/docs/classes/Camera.html), at least in a deprecated way, as its constructor is currently publicly documented.
- [x] Consider exposing [`Sprite`](http://p5play.molleindustria.org/docs/classes/Sprite.html), at least in a deprecated way, as its constructor is currently publicly documented.
- [x] Add/remove/update documentation to reflect any changes in API ~~(deprecated methods, `pInst` as first argument for constructors, etc)~~.


toolness:
Hey @islemaster, would you be interested in code reviewing this at all? It'd be really great to have a second pair of eyes on this, but no worries if not.

islemaster:
I'll be happy to. I can take a look later today.

toolness:
Awesome, thanks!!! I just updated the description of this PR to reflect some API design considerations too.

lib/p5.play.js
```diff
@@ -33,6 +33,92 @@ factory(root['p5']);
...
+//
+//   var pInstBind = createPInstBinder(pInst);
+//
+//   var createVector = pInstBind('createVector');
+//   var loadImage = pInstBind('loadImage');
+//
+// The above will create functions createVector and loadImage, which can be
+// used similar to p5 global mode--however, they're bound to specific p5
+// instances, and can thus be used outside of global mode.
+function createPInstBinder(pInst) {
```
toolness: I'm not sure how I feel about this `createPInstBinder` thing--I admit that I did it partly out of laziness, but also because I didn't want to hamper readability by having a million things be prefixed with `pInst.` or `this.`.

islemaster:
A couple of initial thoughts, more on the way.

1. There's a _lot_ going on here, and a few improvements that look like they could be extracted and merged as separate PRs first.  It could help us reason cleanly about this change.  Candidates:
  * [Add .jshintrc](https://github.com/toolness/p5.play/commit/8cfb08be624e2295908654c04f2df548eccc1c34) and [squelch jshint 'is not defined' errors for window](https://github.com/toolness/p5.play/commit/22d41064354e54af9b67f89302eeeb5633194a97)
  * [Add sinon.js](https://github.com/toolness/p5.play/commit/9a9461969f443e42b5f3cce629ee857acb301601) and [use sinon.stub() in keycodes tests](https://github.com/toolness/p5.play/commit/14574daac675ccab6c270d6c27ee300b056e05db)
  * Maybe [Upgrade p5.js from 0.4.14 to 0.4.21](https://github.com/toolness/p5.play/commit/cb6803dac80fe7491fd94d56b87e376c456bee6f)?  Not sure from your comments what the motivation was behind the upgrade.

2. For all the constructors you're deprecating, did you consider exposing curried versions on the instance that use the original API?  That way client code could just update from `new Group()` to `new p5instance.Group()` (instead of `new Group(p5Instance)`).

Thanks for tackling this big change!

toolness:
Great ideas! I've just filed those PRs, and will implement your suggestion for the curried constructors.

toolness:
Oh, do you have time to take a look at the new PRs? It's ok if not, just let me know.

islemaster:
I do ; sorry, I'm on Pacific time and can review more in about 90 minutes.

toolness:
Awesome, thanks!

So in 768b4145ae025075472c7ac6f30597c1f8b6d12d I added the currying you suggested.

Now I *think* all that's left is to decide whether we also want to expose the non-documented constructors `CircleCollider`, `AABB`, and `Quadtree`, and possibly add more unit tests.

lib/p5.play.js
```diff
@@ -33,6 +33,113 @@ factory(root['p5']);
...
+
+      return value;
+    }
+  });
+}
+
+// This returns a factory function, suitable for passing to
+// defineLazyP5Property, that returns a subclass of the given
+// constructor that is always bound to a particular p5 instance.
+function boundConstructorFactory(constructor) {
```
toolness: Ug, it's a factory for creating constructors... Not a huge fan of the multiple levels of indirection but it's all I could think of for now.

lib/p5.play.js
```diff
@@ -33,6 +33,113 @@ factory(root['p5']);
...
+    var pInst = this;
+
+    function F() {
+      var args = Array.prototype.slice.call(arguments);
+
+      return constructor.apply(this, [pInst].concat(args));
+    }
+    F.prototype = constructor.prototype;
+
+    return F;
```
islemaster: Why not use [Function.prototype.bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) to curry the constructor?

```javascript
return function createBoundConstructor() {
  return constructor.bind(undefined, this);
};
```

See [this jsbin to see this type of curried constructor in action](https://jsbin.com/lutoqogedu/edit?js,console).

Also, I'm not sure there's a reason to name the inner function you're returning.

lib/p5.play.js
```diff
@@ -33,6 +33,113 @@ factory(root['p5']);
...
+// stored here.
+var _globalModeProperties = {};
+
+// This provides a way for us to lazily define properties that
+// are global to p5 instances.
+//
+// Note that this isn't just an optimization: p5 currently provides no
+// way for add-ons to be notified when new p5 instances are created, so
+// lazily creating these properties is the *only* mechanism available
+// to us. For more information, see:
```
islemaster: I'm not sure I grok this; Why do we need to know when new p5 instances are created?  Since we have access to the prototype, don't we have what we need to modify all instances (past and future) already?

I suspect the answer has something to do with global mode, and I need to dig in to how global mode works in vanilla p5 before I get much further in this review.  I suspect you're on the right track here - replacing bare properties (like `allSprites`) with proper getters and setters lets us add those properties to the 'prototype' without having access to the constructor.

lib/p5.play.js
```diff
@@ -33,6 +33,113 @@ factory(root['p5']);
...
+
+      if (this instanceof p5 && !this._isGlobal) {
+        value = getter.call(this);
+
+        // Override our prototype property with an instance property that
+        // contains our lazily-created value.
+        Object.defineProperty(this, name, {
+          enumerable: true,
+          value: value
+        });
```
islemaster: I feel like this goes deeper than necessary.  At this point, `this` is bound to the p5 instance, correct?  Why not just write:

```javascript
this[name] = getter.call(this);
```

Also, I'd rename `getter` to `initialValue`.

toolness: Hmmm, when I do that I get `TypeError: setting a property that has only a getter`. I think I may have tried your suggestion originally but fallen back to `Object.defineProperty` when that error was thrown?

toolness: Good idea about `initialValue`, though that makes it seem like `initialValue` ought to be the actual initial value, rather than a function that returns the initial value--what do you think of `getInitialValue`, or `initialValueGetter`?

lib/p5.play.js
toolness: Oh, actually I don't believe it has anything to do with p5 global mode. It could have something to do with me being a bit rusty with JS, though. :)

Basically I was trying to figure out how to deal with e.g. the following situation:

```js
var pInst = new p5(function(sketch) {
  sketch.setup = function() {
    console.log(sketch.allSprites);
  };
});
```

It seems like the `allSprites` that's logged to the console should be one that's specific to the `sketch` instance, and not e.g. one that's inherited from `p5.prototype`. (For example, if there were a destructive method on `Group` like, say, `setOpacity()`, we wouldn't want `allSprites.setOpacity(0.5)` in the first line of our `setup()` function to actually set the opacity of all sprites on `p5.prototype`, right?)

But I guess I don't know of a way to do this without using something like `defineLazyP5Property`, or having some mechanism to be notified of new instance construction like https://github.com/processing/p5.js/issues/1263.

lib/p5.play.js
toolness: Ok, so this took me way longer than it should have to figure out, but the problem with using [`bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) here is that it actually prevents the bound constructor from being used as a superclass--which is actually required right now because that bound constructor is being passed to `construct()`, which uses it as a superclass (which itself is kind of gross, but I guess addressing that is a different conversation).

While the `thisArg` you're passing to `bind()` is being ignored when using the `new` operator, it *isn't* being ignored when using [`apply()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) on it--in fact, whatever was passed as the first argument of `apply()` is being thrown out in favor of the `thisArg` previously passed to `bind()`, which is `undefined`. In non-strict mode, that gets converted to the global object, which means that all `apply()` calls are applied to `window`. :sob: 

[I've added an example of this to your jsbin example here.](https://jsbin.com/pagete/edit?js,console) It basically creates a trivial subclass of a bound constructor, and shows that when using `apply()` to invoke the bound constructor on the subclass instance, the bound constructor is instead invoked on `window`.

> Also, I'm not sure there's a reason to name the inner function you're returning.

Oh, I just did that to (hopefully) improve readability. I tend to get confused by these multiple levels of indirection where we have functions returning factories that create constructors, so I thought naming the inner function might help people understand it better. If you think it's confusing, though, I can remove it.

islemaster: Nice detective work! :grin: Okay, totally fair, considering we're doing horrendous things to these constructors both up and down the chain.

I'm cool with leaving the named inner function named, just not used to seeing it that way.  I had a coworker point out that naming otherwise anonymous functions can be handy if you ever call `toString` on them, too.

lib/p5.play.js
islemaster: Okay, after studying how p5 initializes (global mode in particular) and the amount of initialization it's doing in its constructor (ick) I'm more comfortable with this approach.  It sort of looks like p5 doesn't want any public instance properties, so creating these properties with getters and setters on the prototype might be the best approach.

There's a slight change in behavior here, but it might be a useful one: Your inner call to `Object.defineProperty` (line 64) is explicitly creating a read-only property on the instance, because `writable` defaults to false ([see MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)).  It might be worth naming this `defineLazyReadOnlyP5Property` to call that out.

Here's what I'm wondering: Should we follow [this pattern in p5](https://github.com/processing/p5.js/blob/master/src/core/core.js#L286) and unify the instance and global paths?  Something like this:

```javascript
function defineLazyReadOnlyP5Property(name, getter) {
  Object.defineProperty(p5.prototype, name, {
    configurable: true,
    enumerable: true,
    get: function () {
      var context = (this instanceof p5 && !this._isGlobal) ? this : window;
      if (typeof context._p5PlayProperties === 'undefined') {
        context._p5PlayProperties = {};
      }
      if (!(name in context._p5PlayProperties)) {
        context._p5PlayProperties[name] = getter.call(context);
      }
      return context._p5PlayProperties[name];
    }
  });
}
```

lib/p5.play.js
islemaster: This might be silly, but I might make "non-lazy" version that lets you just pass a value, not a function; just for convenience.  Then you could do this in your cases that don't require a reference to the instance/context:

```javascript
defineP5Property('allSprites', new Group());
```

lib/p5.play.js
islemaster: What am I saying?  You can close over the value of the property, no need for the cache object!

```javascript
function defineLazyReadOnlyP5Property(name, getter) {
  var propertyValue, everAccessed = false;
  Object.defineProperty(p5.prototype, name, {
    configurable: true,
    enumerable: true,
    get: function () {
      var context = (this instanceof p5 && !this._isGlobal) ? this : window;
      if (!everAccessed) {
        propertyValue = getter.call(context);
        everAccessed = true;
      }
      return evaluatedGetter;
    }
  });
}
```

I think you could even do a read/write variant this way:

```javascript
function defineLazyP5Property(name, initialGetter) {
  var propertyValue, everAccessed = false;
  Object.defineProperty(p5.prototype, name, {
    configurable: true,
    enumerable: true,
    writable: true,
    get: function () {
      var context = (this instanceof p5 && !this._isGlobal) ? this : window;
      if (!everAccessed) {
        propertyValue = getter.call(context);
        everAccessed = true;
      }
      return evaluatedGetter;
    },
    set: function (newValue) {
      propertyValue = newValue;
      everAccessed = true;
    }
  });
}
```

Stop me if this gets annoying :grin:

lib/p5.play.js
```diff
@@ -42,7 +149,10 @@ factory(root['p5']);
 * @property allSprites
 * @type {Group}
 */
-p5.prototype.allSprites = new Group();
+
+defineLazyP5Property('allSprites', function() { return new Group(); });
+
+p5.prototype.spriteUpdate = true;
```
islemaster: Isn't this a problem - the whole reason you wrote `defineLazyP5Property`?  I assume the value of `spriteUpdate` isn't shared across all p5 instances.  If this is working, why can't you just
```javascript
p5.prototype.allSprites = new Group():
```
since `Group` has no dependencies on p5 either?

islemaster: It might be easiest to answer these questions by writing some tests that verify these properties are not shared when you make multiple instances.

lib/p5.play.js
```diff
@@ -701,8 +800,37 @@ p5.prototype._ensureInGlobalMode = function() {
...
+  var line = pInstBind('line');
+  var noFill = pInstBind('noFill');
+  var fill = pInstBind('fill');
+  var textAlign = pInstBind('textAlign');
+  var textSize = pInstBind('textSize');
+  var text = pInstBind('text');
+  var rect = pInstBind('rect');
+  var cos = pInstBind('cos');
+  var sin = pInstBind('sin');
+  var atan2 = pInstBind('atan2');
```
islemaster: This is basically an alias for the following, right?
```javascript
var atan2 = pInst.atan2.bind(pInst);
```

I'm fine with it, just making sure I'm reading this right.

islemaster:
Recommendation: Also break the following commit out into its own PR:

* [f1363d1 - repo now contains a single version of p5.play.js.](https://github.com/toolness/p5.play/commit/f1363d146de27dd8a5f7f2b91fa1c2525788ff65)

lib/p5.play.js
toolness: Hehe :) Sure, I think your first code example [here](https://github.com/molleindustria/p5.play/pull/22#discussion_r53377780), of using the `context` pattern from p5, is a good one! Is that the one you're thinking we should use? It seems your following suggestions have ~~strikethrough~~ before them, so just wanted to make sure.

lib/p5.play.js
toolness: Oh, that particular example you gave would actually break things, wouldn't it? I think it'd mean that the same `Group` instance would be used across all instances, which would mean that adding a sprite to one would add it to the others too...

lib/p5.play.js
toolness: Oh good idea about the tests! I'll do that.

So the main reason I wrote `defineLazyP5Property` was actually so that *mutable* properties, like `allSprites`, would be per-instance, rather than global. e.g. so that calling `sketch1.allSprites.add(mySprite)` wouldn't also add the sprite to `sketch2.allSprites`. Since booleans like `spriteUpdate` are immutable, and since assigning to the property will actually set it on a particular p5 instance rather than the prototype, I thought it would be OK to put the "starting" value on the prototype. Does that make sense?

In any case, I totally agree that tests would serve both as good executable documentation and as protection against regressions, so I'll do that!

lib/p5.play.js
toolness: Yup!

toolness:
Good idea about splitting `f1363d146de27dd8a5f7f2b91fa1c2525788ff65` into its own PR, I will do that!

lib/p5.play.js
islemaster: That's true.  Sorry, I'm finally wrapping my head around this.

lib/p5.play.js
islemaster: Right.  The ones with strikethrough would break things :dizzy_face: :palm_tree: .

lib/p5.play.js
islemaster: Yep, that make sense now.  Thanks for clarifying!

islemaster:
After all that, I don't think I have anything else to add.  Thanks for your patience with my gradual understanding of the initialization challenges around p5 instance properties!

lib/p5.play.js
GoToLoop: No need to **bind()** **atan2()** or any ***Math*** utility function, for all of them are static stateless and doesn't use `this` internally.

islemaster: The `atan2()` method being bound here isn't `Math.atan2` but [`p5.prototype.atan2`](https://github.com/processing/p5.js/blob/1dedc4236aeab2312dfd4e1e5460e06081ce7628/src/math/trigonometry.js#L162) which checks `this._angleMode` to do automatic conversion from degrees to radians in some cases.

GoToLoop: Oops! Forgot about *_angleMode*! Which of course is a state.  *_*

islemaster: :+1: No worries!  I did a double-take the first time I saw that too.

toolness: Hehe, as did I! 

That said, I'm wondering if, in the future, perhaps we should actually refactor this code to just use the analogous `Math` functions instead.  Aside from simplifying the code, it might also protect the library from weird unexpected state changes--for example, I have no idea if p5.play still works properly if the user's sketch decides to fiddle with [`angleMode()`](http://p5js.org/reference/#/p5/angleMode) :)

toolness:
Awesome, thanks @islemaster! And no worries, I'm actually learning how this code works myself too, hehe. It's been very helpful to have a second set of eyes reading over this.

PR #26, opened 2016-02-09
Light documentation and code cleanup

islemaster:
I was reviewing the library today and spotted some low-hanging fruit fixes, so I ran with it :grin:.

* The collection of keycodes had `MINUS` misspelled as `MINUT`
* The `friction` property of `Sprite` had a documented default of -1, but it's real default was 1.
* The `removed` property of `Sprite` said "False when the sprite has been removed" which is the opposite of what it means.
* There was a great deal of duplicate code in the key and mouse state queries, so I extracted it.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -226,9 +226,16 @@ p5.prototype.animation = function(anim, x, y) {
 //variable to detect instant presses
 var keyStates = {};
 var mouseStates = {};
+
+/** @typedef {number} KeyState */
+
+/** @type {KeyState} */
 var KEY_IS_UP = 0;
+/** @type {KeyState} */
 var KEY_WENT_DOWN = 1;
+/** @type {KeyState} */
 var KEY_IS_DOWN = 2;
+/** @type {KeyState} */
 var KEY_WENT_UP = 3;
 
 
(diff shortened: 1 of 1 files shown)


toolness:
Awesome, thanks for the PR!

Would you mind actually splitting this up into a few separate PRs? I'd like to integrate the pure documentation changes of existing functionality immediately (a004051ae4bab3200bc75895e13efbd50374706f and fdc879e58a321433a1e6705552a4b097ec9c2f4e), while the other changes might require a bit more thinking:

* I believe that `KEY_IS_UP` etc are currently *private* to p5.play's implementation, and are not surfaced to end-users. Exposing them could make the library feel more complex and overwhelming to newcomers, but that's really up to @molleindustria to decide. I think the safest fix may be to simply make `isKeyInState` a private implementation function rather than a public one.
* If we want to expose those constants, we'll have to add them to `p5.prototype` I think, otherwise they'll be inaccessible as they're defined in a closure.
* `MINUT` -> `MINUS` is a great typo fix but I'm wondering if we should still provide a deprecated `MINUT` (perhaps with a deprecation warning logged) so that existing code which relies on p5.play doesn't break immediately.

islemaster:
No problem, thanks for the feedback.  I'll open separated PRs later today:

* One with only documentation changes
* One with the `MINUT` -> `MINUS` fix, leaving `MINUT` in with a deprecation warning.
* One with the method extractions.

PR #27, opened 2016-02-11
Documentation fixes

islemaster:
Broken out from PR #26:

* The friction property of Sprite had a documented default of -1, but it's real default was 1.
* The removed property of Sprite said "False when the sprite has been removed" which is the opposite of what it means.


toolness:
Awesome, thanks!

PR #28, opened 2016-02-11
Deduplicate key- and mouse-state checks

islemaster:
Broken out from PR #26:

Extract common code from the methods that check for key state or mouse button state.

Per your request, I've marked the extracted helper methods as `@private`.  I've also added a leading underscore; I'm not sure what the convention is in this repo; [this method](https://github.com/molleindustria/p5.play/blob/master/lib/p5.play.js#L3523) was the only example I could find of a private method on a prototype.

If you'd prefer that I take these helpers off the prototype entirely I can do so, but that seems like a step away from [supporting instance mode](https://github.com/molleindustria/p5.play/issues/12) since they do depend on some state that's currently global.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -321,9 +321,15 @@ defineLazyP5Property('_p5play', function() {
   };
 });
 
+/** @typedef {number} KeyState */
+
+/** @type {KeyState} */
 var KEY_IS_UP = 0;
+/** @type {KeyState} */
 var KEY_WENT_DOWN = 1;
+/** @type {KeyState} */
 var KEY_IS_DOWN = 2;
+/** @type {KeyState} */
 var KEY_WENT_UP = 3;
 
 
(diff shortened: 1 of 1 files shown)


islemaster:
Just caught and fixed a bug with this PR, inherited from the original implementation of two out of three key handlers:

`keyWentDown()` was doing this correctly, calling `keyIsDown(keyCode)` with the mapped numeric keycode.

`keyWentUp()` and `keyDown()` were calling `keyIsDown(key)`, passing the unmapped argument which, in some cases, would be a string.  This would initialize the key state incorrectly.

Fortunately this wasn't breaking often, since most keystate updates happen in readPresses, which did it right.

islemaster:
Note: I have a set of tests for key mappings to be included with the change that deprecates some bad key aliases.

islemaster:
Rebased after merge of #31.  I think this is back to being ready for review.

toolness:
Er, so now that we finally landed #22, this needs to be rebased/remerged... If that's a hassle, let me know and I'll try to fix it up!

islemaster:
On it :grin:!

islemaster:
Rebased after merge of #22.  Take another look?

lib/p5.play.js
```diff
@@ -321,9 +321,15 @@ defineLazyP5Property('_p5play', function() {
   };
 });
 
+/** @typedef {number} KeyState */
+
+/** @type {KeyState} */
```
toolness: I'm curious if these `@typedef` and `@type` comments will cause this documentation to show up in the reference docs for users of p5.play? Since it's private to the implementation of p5.play, that could be confusing for them... I guess we won't find out until we fix #41! :grin: 

In the meantime though, I'm happy just accepting this as-is.

toolness:
This looks great @islemaster! Sorry it took so long to merge.

lib/p5.play.js
islemaster: I can experiment with this and open a PR adding a yuidoc task.

PR #29, opened 2016-02-16
Fix 'SPACE' and ' ' key aliases.

islemaster:
The key aliases `'SPACE'` and `' '` were mapped to key code 33, which actually means <kbd>PgUp</kbd> (you can see `'PAGE_UP'` mapped to 33 in the diff).   This change maps them to code 32 which is actually the <kbd>Spacebar</kbd>.

Workaround prior to this change was just passing key code 32 to the `keyWentUp()`, `keyWentDown()`, etc. checks, instead of a string alias.

I propose that there's no useful backwards compatibility to provide here, since it's unlikely that somebody wrote a game using the `'SPACE'` alias, figured out that it mapped to <kbd>PgUp</kbd>, and decided they were satisfied with that.


toolness:
That makes sense to me, thanks for the fix!

What platforms have you tried this out on? Just want to be fairly confident that it's indeed the correct code for all relevant platforms/locales/etc before merging.

islemaster:
I'm developing on Ubuntu 15.04, Chrome 48, and haven't testing this particular change on other platforms.  However in this case, I think documentation others have done ought to be sufficient.  <kbd>Space</kbd> has been keycode 32 (hex 0x20) since [RFC 20, 1969](https://tools.ietf.org/html/rfc20).

For a brief overview of actual browser support, see [MDN's `keyCode` documentation](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode):

screenshot from 2016-02-16 14-06-44

toolness:
Oh nice, that's perfect, thanks!

PR #30, opened 2016-02-16
Add shorter aliases for arrow keys

islemaster:
Proposal: In addition to the `'UP_ARROW'`, `'DOWN_ARROW`', etc. aliases currently available for the arrow keys, support `'UP'`, `'DOWN'`, `'LEFT'` and `'RIGHT'` aliases which are easier to guess than their snake_case counterparts.


toolness:
That makes sense to me, thanks for the fix!

PR #31, opened 2016-02-16
Deprecate bad key aliases

islemaster:
Broken out from PR #26:

There are some misspelled key aliases in the `KEY` property:

* `MINUT` instead of `MINUS`
* `COMA` instead of `COMMA`

This change corrects the typos, and also introduces a 'deprecated alias' list.  Deprecated aliases are still accepted, but will generate a warning in the browser console when used with a recommendation to change to a corrected alias.

> Key literal "MINUT" is deprecated and may be removed in a future version of p5.play. Please use "MINUS" instead.

This also introduces a set of tests for the new `keyCodeFromAlias_` private lookup helper:

```
  keyCodeFromAlias_
    key aliases
      ✓ maps every alias to a numeric keycode 
      ✓ is case-insensitive 
      ✓ does not warn when looking up a regular alias 
      ✓ maps MINUS to 109 
      ✓ maps COMMA to 188 
    deprecated aliases
      ✓ maps every deprecated alias to a valid key alias 
      ✓ does not include any deprecated aliases in key aliases 
      ✓ aliases 'MINUT' to 'MINUS' 
      ✓ warns when using MINUT 
      ✓ aliases 'COMA' to 'COMMA' 
      ✓ warns when using COMA 
```

Note: I've set up this PR to be independent of #28, but they do have some overlap so I'll need to resolve a conflict on whichever PR we merge second.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -305,9 +305,13 @@ p5.prototype.keyDown = function(key) {
   var keyCode;
 
   if(typeof key == "string")
-    keyCode = KEY[key.toUpperCase()];
+  {
+    keyCode = this.keyCodeFromAlias_(key);
+  }
   else
+  {
     keyCode = key;
+  }
 
   //if undefined start checking it
   if(keyStates[keyCode]===undefined)
@@ -436,9 +440,8 @@ p5.prototype.mouseWentDown = function(buttonCode) {
 * Key.tab = 9
 *
 * @property KEY
-* @type {Group}
+* @type {Object}
 */
-
 p5.prototype.KEY = {
     'BACKSPACE': 8,
     'TAB': 9,
@@ -509,7 +512,7 @@ p5.prototype.KEY = {
     '9NUMPAD': 105,
     'MULTIPLY': 106,
     'PLUS': 107,
-    'MINUT': 109,
+    'MINUS': 109,
     'DOT': 110,
     'SLASH1': 111,
     'F1': 112,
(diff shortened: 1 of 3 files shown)


toolness:
Oh nice!! Thanks so much for the tests... I'll take a closer look at this in the next day or two.

lib/p5.play.js
```diff
@@ -3990,3 +4027,23 @@ p5.prototype.registerMethod('post', cameraPop);
 //deltaTime
 //p5.prototype.registerMethod('pre', updateDelta);
 
+/**
+ * Log a warning message to the host console, using native `console.warn`
+ * if it is available but falling back on `console.log` if not.  If no
+ * console is availble, this method will fail silently.
```
toolness: Very minor typo, `availble` -> `available`.

lib/p5.play.js
```diff
@@ -3990,3 +4027,23 @@ p5.prototype.registerMethod('post', cameraPop);
...
+ * @method warn
+ * @param {!string} message
+ */
+p5.prototype.warn_ = function (message) {
```
toolness: Hmm, I guess my only concern here is that it might conflict with other p5 add-ons that might also define a `warn_` function? (Actually, I might be contributing such a function to p5 itself soon.)

Would it make more sense to have this be a private `warn_` function defined within the immediately executing function that encloses p5.play's source? Perhaps we could expose such functions to tests via a `p5.prototype._p5play` object that contains them, or somesuch?

I'm not really sure what the best solution is, curious if you have any ideas.

lib/p5.play.js
```diff
@@ -525,12 +528,46 @@ p5.prototype.KEY = {
...
+ *          if no key code matching the given alias is found.
+ * @private
+ */
+p5.prototype.keyCodeFromAlias_ = function (alias) {
+  alias = alias.toUpperCase();
+  if (this.KEY_DEPRECATIONS[alias]) {
+    this.warn_('Key literal "' + alias + '" is deprecated and may be removed ' +
+      'in a future version of p5.play. ' +
+      'Please use "' + this.KEY_DEPRECATIONS[alias] + '" instead.');
+    alias = this.KEY_DEPRECATIONS[alias];
```
toolness: Nice, I like these deprecation warnings!

test/unit/keycodes.js
```diff
@@ -0,0 +1,99 @@
+describe('keyCodeFromAlias_', function() {
+  var lastConsoleMessage, originalWarnFunction;
+
+  beforeEach(function () {
```
toolness: YAY TESTS :grin:

lib/p5.play.js
toolness: Just fixed this one myself.

toolness:
This looks great, thanks!

lib/p5.play.js
islemaster: I wondered about this too; I always feel bad writing yet another log wrapper.  In this case, though, it doesn't just add a guard against missing browsers, it's a useful abstraction to allow testing the warnings.  For that reason I don't want it to be totally private.  In a perfect world, I'd want to inject a logging interface when I construct p5.play, but without instance mode that's tricky right now.

PR #32, opened 2016-02-17
Follow convention for private method naming and annotation

islemaster:
I realized that in the last few changes, I haven't been consistent with the new private methods I've introduced.  This change is a pure refactor that brings things in line.

**My understanding of the convention for private methods:**

* Private methods are named with a `_leadingUnderscore`
* Private methods are JSDoc annotated as [`@private`](http://usejsdoc.org/tags-private.html)

**Fixes made:**

* Rename `keyCodeFromAlias_` to `_keyCodeFromAlias`
* Rename `warn_` to `_warn`
* Correct JSDoc for `_warn`: Had wrong method name and was missing `@private` annotation
* Add JSDoc for `_ensureInGlobalMode`

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -245,7 +245,7 @@ p5.prototype.keyWentDown = function(key) {
   var keyCode;
 
   if(typeof key == "string")
-    keyCode = this.keyCodeFromAlias_(key);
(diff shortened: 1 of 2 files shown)


toolness:
Thanks!

Issue #18, opened 2015-12-16
overlapPixel example not working


mrjoshida:
In the [point and pixel overlap example](http://p5play.molleindustria.org/examples/index.html?fileName=collisions5.js#) it looks like overlapPixel is not working - the poor triangle falls straight through the platform.

molleindustria:
Yeah I noticed that, I think something in p5's get() changed and broke it. I'll check it as soon as I can

mrjoshida:
Looks like it's been fixed in p5.js https://github.com/processing/p5.js/issues/342 , pulling an updated version fixes the example

toolness:
Thanks @mrjoshida! It looks like the p5.play website now uses `p5.js v0.4.21 January 04, 2016`, which should include that fix, as the triangle is no longer falling straight through the platform. I'm going to close this issue now, but feel free to re-open it if you disagree.

PR #33, opened 2016-02-17
Upgrade p5.js from 0.4.14 to 0.4.21.

toolness:
This upgrades p5.js, used by the examples and test suite, from [0.4.14](https://github.com/processing/p5.js/releases/tag/0.4.14) to [0.4.21](https://github.com/processing/p5.js/releases/tag/0.4.21).

In particular, upgrading fixes the following bugs for us:

* The point and pixel overlap example (#18). This was fixed by https://github.com/processing/p5.js/issues/342.
* `p5.prototype.remove()` throwing an exception (needed by the teardown of tests for #22). This was fixed by https://github.com/processing/p5.js/pull/1054.

I think there may have been a few more bugs that this fixes for #22 but the above should be enough to justify the upgrade.


islemaster:
:+1: Awesome that this is just a drop-in replacement!

toolness:
Woot thanks!

PR #34, opened 2016-02-17
Use sinon.js for stubbing and spying in tests

toolness:
This simplifies the existing keycodes test suite a bit, and will make it easier to stub out more tests in the future (particularly the tests already in #22).


islemaster:
:+1: Awesome!  This is much nicer.

toolness:
Thanks!

PR #35, opened 2016-02-17
Add support for jshint

toolness:
This adds support for [jshint](http://jshint.com/). It doesn't fix the vast majority of issues, which we can address in other PRs (in particular, the rest of the `**** is not defined` type warnings are fixed in #22).

Once we've fixed all the warnings, we should probably run `jshint` as part of the test suite, to ensure that warnings don't creep back into the codebase.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,11 @@
+{
+  "eqnull": true,
+  "asi": true,
+  "undef": true,
+  "globals": {
+    "p5": true,
+    "define": false,
+    "require": false
+  },
+  "maxerr": 1000000
+}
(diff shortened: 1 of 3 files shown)


islemaster:
Recommendation: Add the following to package.json under `"scripts"`:

```
"lint": "jshint lib/*.js test/unit/*.js"
```

This way you can run linting locally using `npm run lint`, and bare `npm run` (which lists available scripts) will reveal that linting is available.

islemaster:
In lieu of (or maybe in addition to) making linting part of the test suite, consider integrating [Hound](https://houndci.com/) for linting checks on all pull requests.

toolness:
Great ideas! I'll add the `npm run lint` and then merge this.

Issue #25, opened 2016-02-03
Forge a release process


toolness:
I'm not sure what the release process is for p5.play, but we should document it so folks can continue to maintain it. In #24 a bug that was fixed over two months ago was re-reported, presumably because a new version of p5 with the bugfix hadn't been released yet.

For reference, [here's a release process I instituted](https://github.com/mozilla/teach.mozilla.org/blob/develop/RELEASE.md) for a previous project I worked on.

We should also probably make the official project page at p5play.molleindustria.org something that other maintainers can deploy to as part of the release process--perhaps either through a Travis CI hook or a gh-pages branch.

PR #36, opened 2016-02-21
The repository should only contain one copy of p5.play

toolness:
This gets rid of the extra copy of `p5.play.js` so that we don't need to pollute the repository's commit history by constantly updating two separate copies of the library.

However, doing this means `examples/index.html` now loads `../lib/p5.play.js`, which means that the `examples` directory can't simply be copied somewhere on its own. This will likely mean some changes to our [release process](https://github.com/molleindustria/p5.play/issues/25). (To the extent that we even *have* a release process yet.)


islemaster:
:+1:

toolness:
Thanks!

PR #38, opened 2016-02-21
Smoke test examples on all browsers.

### LINKED ISSUE #20: Smoke test examples as part of test suite
A bit similarly to https://github.com/processing/p5.js/pull/1134, it'd be nice to run all the examples in the `examples` directory during `npm test` just to make sure that they don't throw anything during setup or after rendering the first frame.

toolness:
This fixes #20~~, but doesn't currently work on PhantomJS, I think because PhantomJS is basically accessing the stuff over a `file:` protocol, which will make things barf. I guess we should start up a basic static web server for the PhantomJS tests to run against, but I'd like to file that as a separate bug~~.

The implementation works by running each example in an iframe and [registering](https://github.com/processing/p5.js/wiki/Libraries#use-registermethod-to-register-functions-with-p5-that-should-be-called-at-various-times) a `post` callback that waits for 3 frames to be drawn before completing. If any errors were thrown during this period, the test fails.

These tests run nicely on PhantomJS and Microsoft Edge (about 50-100ms per test), not-horribly on Chrome (about 100-200ms per test), but extremely slowly on Firefox (about 1.5-2 seconds per test). :disappointed: 

@islemaster want  to take a look at this?

### DIFF
--- a/test/js/bind.js
+++ b/test/js/bind.js
@@ -1,28 +1,29 @@
 // `Function.prototype.bind` polyfill, included in the test environment because
 // PhantomJS does not implement it natively.
 // Source:
-// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Compatibility
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Polyfill
 if (!Function.prototype.bind) {
-  Function.prototype.bind = function (oThis) {
-    if (typeof this !== "function") {
-      // closest thing possible to the ECMAScript 5 internal IsCallable
-      // function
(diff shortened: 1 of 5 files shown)


test/js/bind.js
```diff
@@ -1,28 +1,29 @@
 // `Function.prototype.bind` polyfill, included in the test environment because
```
toolness: The changes here in our `Function.prototype.bind()` polyfill aren't mine; they were changes made to MDN's polyfill due to the following bug: https://gist.github.com/jacomyal/4b7ae101a1cf6b985c60

islemaster:
Woo!  This is so cool! :shipit: Ship it!

toolness:
Awesome, thanks! :grin:

PR #42, opened 2016-02-27
Add tests for input methods.

toolness:
This adds tests for the input methods in p5.play. I actually used it to ensure that #28 didn't cause regressions. :grin: 

However, I think I *might* have also uncovered a bug. The documentation for `keyDown` states:

```
Like p5 keyIsDown but accepts strings and codes
```

However, the implementation is checking to see if the internal key state is specifically equal to `KEY_IS_DOWN`. The problem is that the internal keystate might actually be `KEY_WENT_DOWN`, in which case the method will retrun `false` but `keyIsDown()` will be `true`!

It's understandable that no one has noticed this, because it just means that the result of `keyDown` will be inconsistent with `keyIsDown` for only the frame in which the key went from up to down.

A parallel problem exists with `mouseDown`/`mouseIsDown`.

@islemaster is this something you think we should change? And does the test suite look  OK overall?

### DIFF
--- a/test/unit/input.js
+++ b/test/unit/input.js
@@ -1,0 +1,118 @@
+describe('Input', function() {
+  var myp5;
+  var LEFT_ARROW, LEFT;
+
+  beforeEach(function() {
+    myp5 = new p5(function() {});
+    LEFT_ARROW = myp5.LEFT_ARROW;
+    LEFT = myp5.LEFT;
+  });
+
+  afterEach(function() {
+    myp5.remove();
+  });
+
+  function mouseup(which) {
+    myp5._setProperty('mouseButton', which);
+    myp5._setProperty('mouseIsPressed', false);
+  }
+
+  function mousedown(which) {
+    myp5._setProperty('mouseButton', which);
+    myp5._setProperty('mouseIsPressed', true);
+  }
+
+  function keydown(which) {
+    myp5._onkeydown({which: which});
+  }
+
+  function keyup(which) {
+    myp5._onkeyup({which: which});
+  }
+
+  function nextFrame() {
+    myp5.readPresses();
+  }
+
+  it('reports mouseDown() properly', function() {
+    mouseup(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseDown(LEFT)).to.be.false;
+
+    mousedown(LEFT);
+    nextFrame();
+
+    // TODO: Is this actually the expected behavior?
+    expect(myp5.mouseDown(LEFT)).to.be.false;
+
+    nextFrame();
+
+    expect(myp5.mouseDown(LEFT)).to.be.true;
+  });
+
+  it('reports mouseWentDown() properly', function() {
+    mouseup(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentDown(LEFT)).to.be.false;
+
+    mousedown(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentDown(LEFT)).to.be.true;
+  });
+
+  it('reports mouseWentUp() properly', function() {
+    mousedown(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentUp(LEFT)).to.be.false;
+
+    mouseup(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentUp(LEFT)).to.be.true;
+  });
+
+  it('reports keyDown() properly', function() {
+    keyup(LEFT_ARROW);
+    nextFrame();
+
+    expect(myp5.keyDown(LEFT_ARROW)).to.be.false;
+
+    keydown(LEFT_ARROW);
+    nextFrame();
+
+    // TODO: Is this actually the expected behavior?
+    expect(myp5.keyDown(LEFT_ARROW)).to.be.false;
+
+    nextFrame();
+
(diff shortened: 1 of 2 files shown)


test/unit/input.js
```diff
@@ -0,0 +1,118 @@
...
+    LEFT = myp5.LEFT;
+  });
+
+  afterEach(function() {
+    myp5.remove();
+  });
+
+  function mouseup(which) {
+    myp5._setProperty('mouseButton', which);
+    myp5._setProperty('mouseIsPressed', false);
```
toolness: Since this is using internal p5 methods, it may break when we upgrade the version of p5.js in the repository.

islemaster:
I knew about this. I definitely want to fix it so keyDown and keyWentDown
are not mutually exclusive states. I figured it wouldn't happen soon
though, since it's technically a breaking change. I also want to check
whether @molleindustria did this on purpose, to circumvent a certain set of
beginner mistakes.

If we had a release process and semver I would bundle this with other
breaking changes in a major version. Since things are still pretty fast and
loose (and I don't think anyone auto-updates this library?) maybe we can
just fix it.

This adds tests for the input methods in p5.play. I actually used it to
ensure that #28 <https://github.com/molleindustria/p5.play/pull/28> didn't
cause regressions. [image: :grin:]

However, I think I *might* have also uncovered a bug. The documentation for
keyDown states:

Like p5 keyIsDown but accepts strings and codes

However, the implementation is checking to see if the internal key state is
specifically equal to KEY_IS_DOWN. The problem is that the internal
keystate might actually be KEY_WENT_DOWN, in which case the method will
retrun false but keyIsDown() will be true!

It's understandable that no one has noticed this, because it just means
that the result of keyDown will be inconsistent with keyIsDown for only the
frame in which the key went from up to down.

A parallel problem exists with mouseDown/mouseIsDown.

@islemaster <https://github.com/islemaster> is this something you think we
should change? And does the test suite look OK overall?
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/molleindustria/p5.play/pull/42
Commit Summary

   - Add tests for input methods.

File Changes

   - *M* test/index.html
   <https://github.com/molleindustria/p5.play/pull/42/files#diff-0> (1)
   - *A* test/unit/input.js
   <https://github.com/molleindustria/p5.play/pull/42/files#diff-1> (118)

Patch Links:

   - https://github.com/molleindustria/p5.play/pull/42.patch
   - https://github.com/molleindustria/p5.play/pull/42.diff

—
Reply to this email directly or view it on GitHub
<https://github.com/molleindustria/p5.play/pull/42>.

test/unit/input.js
```diff
@@ -0,0 +1,118 @@
...
+  it('reports mouseWentDown() properly', function() {
+    mouseup(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentDown(LEFT)).to.be.false;
+
+    mousedown(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentDown(LEFT)).to.be.true;
```
islemaster: I feel like for this test to cover the behavior of `mouseWentDown` properly it needs to run a third frame (as the `mouseDown` test does above) and show that `mouseWentDown` goes back to false.

test/unit/input.js
```diff
@@ -0,0 +1,118 @@
...
+
+  function keyup(which) {
+    myp5._onkeyup({which: which});
+  }
+
+  function nextFrame() {
+    myp5.readPresses();
+  }
+
+  it('reports mouseDown() properly', function() {
```
islemaster: Missing a matching `mouseUp` test, for [this method](https://github.com/molleindustria/p5.play/blob/master/lib/p5.play.js#L431).  Oddly enough, p5.play doesn't have a corresponding `keyUp` method.

toolness: Oops, thanks, I missed that because I wrote the key tests first and then copy-pasted them to be mouse tests. :grin:

test/unit/input.js
toolness: Good idea, just added that in d5cc6fb89abaf12050f49151995b410190ba1995.

toolness:
Ok, it looks like there are enough concerns about the fix for the bug that I'm just spinning it off into #43 and we can deal with it later!

test/unit/input.js
```diff
@@ -0,0 +1,148 @@
...
+
+    expect(myp5.mouseWentDown(LEFT)).to.be.false;
+
+    mousedown(LEFT);
+    nextFrame();
+
+    expect(myp5.mouseWentDown(LEFT)).to.be.true;
+
+    mouseup(LEFT);
+    nextFrame();
```
islemaster: This isn't quite what I meant.  With the expected behavior of `mouseWentDown` the `mouseUp(LEFT)` call here shouldn't be necessary for `mouseWentDown(LEFT)` to be false on the next frame.  I was hoping the tests would illustrate this behavior:

Frame | Event | mouseWentDown | mouseDown
----- | ----- | ------------- | ---------
1 | up | false | false
2 | down | true | false
3 | -- | false | true

toolness: Ahh yes, of course! Sorry about that, I think I just fixed this.

islemaster:
Looks good!

toolness:
Woot thanks!

Issue #41, opened 2016-02-27
Add an npm task to generate the documentation


toolness:
I have no idea how to regenerate the reference documentation for this project.

It would be great if we had an `npm docs` command or something that rebuilt them.

It would be *really* cool if `npm start` also rebuilt the docs whenever changes were made to `lib/p5.play.js`, but I suspect this would require adopting grunt or gulp (I prefer gulp, but we can use grunt if we want to maintain parity with p5.js development).

We might also want to consider removing the built docs from the repository, so that we don't need to spam the commit history with rebuilds of the documentation whenever we change it.

islemaster:
I've got a WIP on my [yuidoc-helper branch](https://github.com/molleindustria/p5.play/compare/master...islemaster:yuidoc-helper).

islemaster:
There are a lot of asks in this one issue, but #45 resolved the ask in the title.  The remaining asks seem to be:

* Work with @molleindustria to update the docs at p5play.molleindustria.org/docs
* Somehow automate docs generation, so that a hosted version somewhere is always current with `master`
* Remove the built docs from the repository (presumably after the above work)
* Provide a watcher for the local docs build (I question the usefulness of this :wink:)

@toolness do you think we should:

* Rename this issue to reflect the remaining asks and clarify what's left?
* Resolve this issue and open new ones for the remaining work?

toolness:
Oh, sorry I didn't reply to this!

Hmm, I think we should resolve this issue and open new ones, because resolving issues is fun, and also because I was half-daydreaming when I wrote this issue :grin:

PR #45, opened 2016-03-14
Add `npm run docs` command

islemaster:
Adds an `npm run docs` task to regenerate the p5.play documentation, per issue #41.

Details:

* Does some cleanup of warnings I found when generating the docs
* Adds a link on the `npm start` local dev start page to the local documentation
* I noticed an issue where private methods show up in the documentation table of contents, even when the option to show private methods is unchecked.  I couldn't figure out how to fix this in the underlying YUIDoc theme, so for now I've checked the "show private methods" option by default to minimize confusion.

Does not update the checked-in documentation; I'll do that in another PR, for clarity.

### DIFF
--- a/index.html
+++ b/index.html
@@ -37,6 +37,8 @@ a:active, a:hover, #reference a:hover {
   *
   <a href="test/">Test Suite</a>
   *
+  <a href="docs/">Reference</a>
+  *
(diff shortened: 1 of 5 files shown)


toolness:
Woot awesome!! Weird about the `private` thing but oh well. Thanks!!

PR #47, opened 2016-03-14
Update checked-in documentation

islemaster:
Just checking in the latest output of `npm run docs`.

There's some discussion on #41 on removing the built-in docs from the repository so we don't do this manually, but for now since I don't have access to http://p5play.molleindustria.org/docs/index.html or a way to automatically update the docs there, making sure the checked-in ones are up-to-date seems like a good idea.


toolness:
Oh, sorry I missed this one earlier--thanks!

PR #50, opened 2016-03-19
Do not double-add sprites to groups (Fixes #37)

### LINKED ISSUE #37: Sprites can be added to groups multiple times
Not sure if this is a big deal, but I noticed it when I accidentally wrote the following code in a test:

```js
var mySprite = createSprite(1, 1);
allSprites.add(mySprite);

console.log(allSprites.length);
```

The number `2` will be logged to the console.

This is because `createSprite()` actually adds `mySprite` to `allSprites` on its own, so the call to `allSprites.add()` actually adds it *again*.

This could lead to weird behavior. For instance, since our current implementation of `Group#remove()` only removes the *first* reference to a sprite, this could result in unexpected behavior if someone's code calls it but it doesn't seem to "work" because another reference to the sprite still exists in the Group.

Not sure what we should do about this. Two options I can think of:

* Use the ES6 [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) class instead of `Array` for Groups, and use a [`Set` polyfill](https://github.com/WebReflection/es6-collections) when it's unavailable.

* Alternatively, if we want to keep as much backwards compatibility as possible, we could just add a check to `Group#add()` that first checks to see if the given sprite already exists in the group... The user could still shoot themselves in the foot in a number of ways since `Group` is also an `Array`, but at least we'd prevent the most common ways of doing so.

islemaster:
Adds some simple tests around `Group.add()` and `Group.remove()`, and fixes the following bugs:

* It's no longer possible to add a sprite to a group more than once through `Group.add()`.  Trying to do so is a silent no-op.  (Fixes #37)
* `Group.remove(sprite)` now removes all copies of a sprite from a group - while this shouldn't happen in normal use, it's still possible to get a sprite copy into a group via standard array methods.
* `Group.remove(sprite)` now removes the sprite's reference back to the group that was added by `Group.add()`.

And had to do the following cleanup along the way:
* `Sprite.remove()` and `Group.removeSprites()` iterate more safely, less likely to skip groups/sprites if the array is modified during iteration.

Group tests after this change:
```
  Group
    ✓ is an array 
    add()
      ✓ adds a sprite to the group 
      ✓ gives the sprite a reference to itself 
      ✓ lets you add different sprites to the group 
      ✓ ignores double addition of a unique sprite 
      ✓ throws if passed something besides a sprite 
    remove()
      ✓ removes a sprite from the group 
      ✓ removes the sprite's reference to itself 
      ✓ returns false if the sprite is not a member of the group 
      ✓ removes all copies of a sprite if more than one happen to be present 
      ✓ throws if passed something besides a sprite 
    removeSprites()
      ✓ should remove all sprites 
```

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -2513,10 +2513,9 @@ function Group() {
   * @method clear
   */
   array.removeSprites = function() {
-    var arrayLength = array.length;
-
-    for(var i = 0; i<arrayLength; i++)
+    while (array.length > 0) {
       array[0].remove();
+    }
   }
 
   /**
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -1548,9 +1548,8 @@ function Sprite(pInst, _x, _y, _w, _h) {
     quadTree.removeObject(this);
 
     //when removed from the "scene" also remove all the references in all the groups
-    for(var i=0; i<this.groups.length; i++)
-    {
-      this.groups[i].remove(this);
+    while (this.groups.length > 0) {
+      this.groups[0].remove(this);
     }
   }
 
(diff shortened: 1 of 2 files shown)


toolness:
This is great, thanks!!

PR #51, opened 2016-03-21
Fix #49: Allow circle collisions to bounce at oblique angles

### LINKED ISSUE #49: Circle collisions offset instead of bounce at oblique angles
The issue:
bounce
Repro:

```javascript
var s, g, ball;

function setup() {
  createCanvas(400, 400);
  g = new Group();
  
  s = createSprite(200, 75, 200, 0);
  s.immovable = true;
  s.debug = true;
  s.setCollider('square');
  g.add(s);
  
  s = createSprite(200, 325, 200, 0);
  s.immovable = true;
  s.debug = true;
  s.setCollider('square');
  g.add(s);
  
  s = createSprite(251, 275, 0, 0);
  s.setCollider('circle', 0, 0, 50);
  s.immovable = true;
  s.debug = true;
  g.add(s);
  
  s = createSprite(149, 125, 0, 0);
  s.setCollider('circle', 0, 0, 50);
  s.immovable = true;
  s.debug = true;
  g.add(s);
  
  ball = createSprite(200, 200, 0, 0);
  ball.setCollider('circle', 0, 0, 10);
  ball.debug = true;
  ball.velocity.y = 5;
}

function draw() {
  background('#fff');
  ball.bounce(g);
  drawSprites();
}
```

joshlory:
Fix for #49.  Implements 2d circle-circle elastic collisions based on this formula from https://en.wikipedia.org/wiki/Elastic_collision#Two-dimensional:

3a70e57f4a5cc0e5e0e11be153aa4b10

Demo here: https://jsfiddle.net/6p5wk8g1/show

---

Before to this change, velocity is only adjusted along one axis:

bounce_before

After this change, oblique bounce angles are possible:

bounce_after

Examples generated with the following repro:
```javascript
var s, g, ball;

function setup() {
  createCanvas(400, 400);
  g = new Group();
  
  s = createSprite(100, 100, 0, 0);
  s.setCollider('circle', 0, 0, 50);
  s.immovable = true;
  s.debug = true;
  g.add(s);
  
  s = createSprite(300, 300, 0, 0);
  s.setCollider('circle', 0, 0, 50);
  s.immovable = true;
  s.debug = true;
  g.add(s);
  
  ball = createSprite(200, 200, 0, 0);
  ball.setCollider('circle', 0, 0, 10);
  ball.debug = true;
  ball.velocity.x = 5;
  ball.velocity.y = 5;
}

function draw() {
  background('#fff');
  ball.bounce(g);
  drawSprites();
}
```

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -2164,6 +2164,27 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
               if(type == "bounce")
               {
+                if (this.collider instanceof CircleCollider && other.collider instanceof CircleCollider) {
+                  var dx1 = p5.Vector.sub(this.position, other.position);
+                  var dx2 = p5.Vector.sub(other.position, this.position);
+                  var magnitude = dx1.magSq();
+                  var totalMass = this.mass + other.mass;
+                  var m1 = 0, m2 = 0;
+                  if (this.immovable) {
+                    m2 = 2;
+                  } else if (other.immovable) {
+                    m1 = 2;
+                  } else {
+                    m1 = 2 * other.mass / totalMass;
+                    m2 = 2 * this.mass / totalMass;
+                  }
+                  var newVel1 = dx1.mult(m1 * p5.Vector.sub(this.velocity, other.velocity).dot(dx1) / magnitude);
+                  var newVel2 = dx2.mult(m2 * p5.Vector.sub(other.velocity, this.velocity).dot(dx2) / magnitude);
+
+                  this.velocity.sub(newVel1.mult(this.restitution));
+                  other.velocity.sub(newVel2.mult(other.restitution));
+                }
+                else {
                 if(other.immovable)
                 {
                   var newVelX1 = -this.velocity.x+other.velocity.x;
@@ -2211,6 +2232,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
                   if(!other.immovable)
                     other.velocity.y = newVelY2*other.restitution;
+                }
                 }
               }
               //else if(type == "collide")


lib/p5.play.js
```diff
@@ -2212,6 +2233,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
                   if(!other.immovable)
                     other.velocity.y = newVelY2*other.restitution;
                 }
+                }
```
joshlory: I intentionally didn't reformat the code in the `else` to minimize the diff. The AABB bounce code will be fixed in a future commit.

lib/p5.play.js
```diff
@@ -2164,6 +2164,27 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
               if(type == "bounce")
               {
+                if (this.collider instanceof CircleCollider && other.collider instanceof CircleCollider) {
+                  var dx1 = p5.Vector.sub(this.position, other.position);
+                  var dx2 = p5.Vector.sub(other.position, this.position);
```
islemaster: I believe `dx2` is guaranteed to be the inverse of `dx1` so you could just do `var dx2 = dx1.mult(-1);`.  It appears to be the faster option ([jsperf](http://jsperf.com/compare-vector-sub-with-mult/2)).

joshlory:
The tunneling code doesn't seem to guarantee that the two circles displace outside each other before calling the bounce code path.  I'll need to update this PR to ensure the circles don't become entangled.

islemaster:
This looks solid to me.  Thanks @joshlory!  Any thoughts about how we could write a unit test around this improved behavior?

@toolness thoughts?  This changes existing behavior, but it seems like an obvious improvement to me.

lib/p5.play.js
joshlory: The problem is that `dx1.mult(-1);` modifies `dx1`, but I already modify `dx1` on [line 2181](https://github.com/molleindustria/p5.play/pull/51/files#diff-84d79f68ce33ce29426e52fd37061b4eR2181).  Copying the vector negates the speed up: http://jsperf.com/compare-vector-sub-with-mult/3.

islemaster:
Here's an expanded example for checking the bounce behavior at various angles, instead of just the diagonal case:

```javascript
var s, g, ball;
var angle = 0;

function setup() {
  createCanvas(400, 400);
  g = new Group();

  s = createSprite(100, 100, 0, 0);
  s.setCollider('circle', 0, 0, 50);
  s.immovable = true;
  s.debug = true;
  g.add(s);

  s = createSprite(300, 300, 0, 0);
  s.setCollider('circle', 0, 0, 50);
  s.immovable = true;
  s.debug = true;
  g.add(s);

  ball = createSprite(200, 200, 0, 0);
  ball.setCollider('circle', 0, 0, 10);
  ball.debug = true;
  ball.velocity.x = 5;
  ball.velocity.y = 5;
  
  positionWalls();
}


function positionWalls() {
    g[0].position.x = Math.sin(angle) * 100 + 200;
    g[0].position.y = Math.cos(angle) * 100 + 200;
    
    g[1].position.x = Math.sin(angle) * -100 + 200;
    g[1].position.y = Math.cos(angle) * -100 + 200;
}

function draw() {
  background('#fff');
  ball.bounce(g);
  drawSprites();
  
  if (ball.position.x > 400 || ball.position.y > 400 || ball.position.x < 0 || ball.position.y < 0) {
      ball.position.x = 200;
      ball.position.y = 200;
      ball.velocity.x = 0;
      ball.velocity.y = 0;
      ball.addSpeed(5, (angle * -180 / p5.prototype.PI) + 90);
  }
  
  if (keyDown('RIGHT')) {
      angle -= .05;
      positionWalls();
  } else if (keyDown('LEFT')) {
      angle += .05;
      positionWalls();
  }
}
```

lib/p5.play.js
islemaster: Nice catch, my bad.

toolness:
Oh cool, I like this!

Since it does technically modify existing behavior, perhaps we should run it by @molleindustria just to make sure it's ok?

I sorta feel like the ideal "test" for this would be to make a "movie" of the second animated gif in the original description for this bug, and then comparing the real behavior of the engine to the movie somehow. It doesn't need to compare every pixel of every frame or anything--perhaps the "movie" could really consist of the position of each sprite during each frame or something...

Hmm. This could be a neat way of testing lots of properties of p5.play that aren't easily expressed as standard mocha-style unit tests, but would definitely require a separate PR. If you think it might be useful, I could try prototyping something...

molleindustria:
Oh I guess I forgot to implement the circle-to-circle bounce? If it works it's cool with me. Thank you all for fixing so many things!

islemaster:
@joshlory and I were chatting about how cool it would be to have a set of integration tests that actually did compare the canvas contents (maybe via [toDataUrl](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL)) over a set number of frames, like a p5.play-specific version of a Selenium or Applitools test.  Then (bonus points!) what if the test exported a GIF comparison of the old/new behavior if the test failed?

(geeking out)

lib/p5.play.js
islemaster: Oh, I see what I was doing wrong in the jsperf - you're using the static `sub` and `mult` functions here, where I was using the instance function (both of which modify the vector in-place).  Did a version comparing the static version properly... and they come out exactly the same. (facepalm)

http://jsperf.com/compare-vector-sub-with-mult/4

toolness:
Hehe, that could be pretty cool--I am a bit worried about false positives but it's probably silly of me. I'll hack something up and get back to you--unless you want to take a stab at it yourself?

islemaster:
All yours - I don't time for it yet.  In the meantime, shall we merge this?

toolness:
Oh sure, sounds good to me!

toolness:
Hmm, so I thought more about the testing idea and realized that the reason I'm worried about doing a pixel-for-pixel comparison of each frame is because I'm not confident that every browser will render to canvas the exact same way. Also, adding a lossless animation to the repository for each test could bloat things quickly, maybe. Hmm.

Aside from that, though, I had a weird idea for recording and replaying sketches like this that I wanted to try out, which you can see here:

http://processing.toolness.org/experiments/recorder/

It basically records all the drawing commands made to p5 during a sketch's execution. We can reuse that recording to create a standalone "movie" sketch that doesn't rely on p5.play, and we can also compare future runs of the sketch against the recording for the purposes of regression testing... And because the recording can itself be expressed as a sketch--albeit a fairly large one, though we can make it smaller by e.g. only recording every 10 frames--we could commit it to the repository and actually possibly have sensible diffs... Hmm. Anyways, just a weird idea, might be too hare-brained.

islemaster:
It's a very cool idea.  Nice work getting a prototype up and running so quick! :tada: 

I wonder if you could simplify it by wrapping the drawing API for the canvas that p5 is drawing to instead of wrapping (almost) every p5 function.  It seems to me if we're going to write an integration test like this, we want to capture 'output' as close to the browser as possible before we hit cross-browser rendering differences (or if it was me, I'd just have written these tests requiring that they run under a particular version of phantom).

I suppose the downsides to that approach are:

1. Recorded output that's 'further' from the p5 code you write, so potentially harder to debug problems when they occur.
2. More verbose output (I assume many p5 functions encapsulate multiple canvas operations)

On the other hand, your 'recording' format would be independent of p5.js, which feels cleaner in a scenario where we're _testing_ p5.js (or p5.play, which seems sufficiently entangled with p5.js that it still feels like a benefit to me).

PR #52, opened 2016-03-21
Fix incorrect rendering by initializing constants later

### LINKED ISSUE #48: Sprite and collider box are drawn off center
Issue: In latest release of p5.play sprite and collider box are drawn off center.

Here's an image of a sprite with debug turned on. The center is drawn correctly but the sprite itself is drawn as if rectMode is set to "CORNER" rather than "CENTER":
collidertest_latest_release

Here's the same sketch using the same version of p5.play that's used in the examples on the your website:
collidertest_using_web_examples_release

You can download the sketch used to create these images here:
[colliderTest.zip](https://github.com/molleindustria/p5.play/files/180124/colliderTest.zip)

You can recreate for yourself by downloading the latest p5.play release from the website.

islemaster:
@cpirich fixed #48 downstream in code-dot-org/code-dot-org#7415.  I've manually extracted just the fix to when constants are initialized, and added a simple unit test to cover the issue.

The test reaches further into p5 rendering details than I'd like, but it checks the misinitialized `CENTER` constant right where it's used, and would complain if anything else changed the default coordinate mode for Sprite rendering, so it seems useful.

@gtoast I used the repro sketch you provided to verify that this fixes the issue you reported.  Please take a look and let me know if you see any problem with this.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -107,13 +107,6 @@ function createPInstBinder(pInst) {
   };
 }
 
-// These are p5 constants that we'd like easy access to.
-var RGB = p5.prototype.RGB;
-var CENTER = p5.prototype.CENTER;
-var LEFT = p5.prototype.LEFT;
-var BOTTOM = p5.prototype.BOTTOM;
-var PI = p5.prototype.PI;
-
 // These are utility p5 functions that don't depend on p5 instance state in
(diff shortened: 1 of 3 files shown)


toolness:
This looks great, thanks for the fix!!

PR #55, opened 2016-03-22
Fix #54: Don't leak render state in Sprite.debug

### LINKED ISSUE #54: Sprite.debug drawing leaks render state
The debug drawing code found in [Spirte.display() at p5.play.js:1485](https://github.com/molleindustria/p5.play/blob/master/lib/p5.play.js#L1485) does not properly push/pop p5 render state, so it can affect rendering operations that take place after it runs.

# Example
An easy example of this is that text rendering breaks when it follows sprite rendering with debug=true, including on subsequent frames.  Here's a small repro case where debug=true when the mouse is down.

p5-debug-bug

You can run this example at http://p5play.molleindustria.org/examples/ by pasting the following code:
```javascript
var circle;

function setup() {
  createCanvas(800,400);  
  
  //create 4 sprites
  circle = createSprite(400, 200);
  //compact way to add an image
  circle.addImage(loadImage("assets/plain_circle.png"));
  
}

function draw() {
  background(255,255,255);  
  

  //if debug is set to true bounding boxes, centers and depths are visualized
  circle.debug = mouseIsPressed;
  
  drawSprites();
  
  textAlign(CENTER, CENTER);
  text('Here is some example text', 0, 0, 800, 200);
}
```

# Related issue
There's a related problem, equally easy to fix, where debug also doesn't draw if the user calls `strokeWeight(0)`.

@joshlory has volunteered to fix these two issues; assigning to him right away.

joshlory:
Fixes #54.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -1484,8 +1484,10 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
       if(this.debug)
       {
+        push();
         //draw the anchor point
         stroke(0,255,0);
+        strokeWeight(1);
         line(this.position.x-10, this.position.y, this.position.x+10, this.position.y);
         line(this.position.x, this.position.y-10, this.position.x, this.position.y+10);
         noFill();
@@ -1505,6 +1507,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
         {
           this.collider.draw();
         }
+        pop();
       }
 
     }


islemaster:
It's definitely an improvment, but in my example case I'm still seeing a stroke/fill change after setting debug=true.  It seems this is more complicated than I thought.

p5-debug-bug-v2

islemaster:
I suspect this last fill/stroke issue is actually a bug in p5, not in p5.play.  Notice that [the `push()/pop()` methods](https://github.com/processing/p5.js/blob/a385d1ca7a1232b0e4d20066a395706ef88d0e7c/src/core/structure.js#L161) don't save or restore the `_fillSet` or `_strokeSet` properties [that get changed when you call `fill`](https://github.com/processing/p5.js/blob/f56df5deab1df1ae9d104ded71e47e1a8cca0ced/src/color/setting.js#L373) or `stroke`.  A quick experiment with the example case shows these properties failing to reset:

screenshot from 2016-03-23 16-41-50

...and they [absolutely have an effect on text rendering.](https://github.com/processing/p5.js/blob/20289b05e7df453c4e809320c7356b0f5e33f0bb/src/core/p5.Renderer2D.js#L1143)  Modifying the `push()` method in p5.js locally appears to fix the problem, but I'm not totally sure what the other ramifications of that change are.  I'll open a bug against p5.js and when it is fixed, and the version we depend on here is fixed, I think we can resolve #54.

islemaster:
Issue opened against p5.js.  That said, I think the changes to p5.play made here are correct, and assuming the problem with p5 gets fixed, will resolve the issue in question.  It's also a definite improvement in the interim.

PR #59, opened 2016-03-27
Fix instance mode breakage when debug mode is active

toolness:
It looks like we added a call to `strokeWeight()` in #54, but we didn't bind it to `pInst`, which meant that sketches in instance mode would crash because it wasn't defined globally.


islemaster:
Oh man, sorry about that.  I suppose a unit test around debug drawing would have caught this, since we're using instance mode in our tests.

Issue #63, opened 2016-03-28
Mark Sprite constructor as @private in docs?


toolness:
I was just reading through the docs and noticed that our `Sprite` constructor is documented as a public function.

However, another method we have to create sprites is the `createSprite` function, which actually seems preferable, because it adds the newly-created spite to `allSprites`. Should we actually deprecate using the `Sprite` constructor directly, or mention in its documentation that the reader will probably want to use `createSprite` instead?

islemaster:
Yes, I suspect we should at least document it as such.  I'm not sure it's worth making it entirely inaccessible though.

Issue #64, opened 2016-03-28
Group.bounce, .displace, .overlap, .collide have incorrect documentation


toolness:
There's a number of problems with the documentation for these methods:

* They claim that a `Group` must be passed in as the first argument; however, as far as I can tell, these methods are all analogous to the eponymous methods in `Sprite`, which actually accept a `Group` *or* a `Sprite` as their first argument.

* Similar to #23, `Group` has two `overlap` methods documented, and no `collide` method documented.

* The examples for these methods seem to be copy-pasted from their eponymous methods in `Sprite`, which is confusing. For example, the `Group.displace()` example says:

  ```js
  sprite.displace(otherSprite, explosion)
  ```

  but it would make more sense to have e.g.

  ```js
  group.displace(otherSprite, explosion)
  ```

PR #66, opened 2016-03-29
Group docs fixes

### LINKED ISSUE #23: Group has two clear methods in API, removeSprites not listed
The API lists two clear methods. The code has @method clear on the removeSprites method. clear API
clear documentation

toolness:
This fixes #23 and #64. It doesn't regenerate the documentation, though, as I didn't want to spam the commit history... :disappointed:

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -2536,7 +2536,7 @@ function Group() {
   * Removes all the sprites in the group
   * from the scene.
   *
-  * @method clear
+  * @method removeSprites
   */
   array.removeSprites = function() {
     while (array.length > 0) {
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -2691,19 +2691,19 @@ function Group() {
   }
 
   /**
-  * Checks if the the group is overlapping another group.
+  * Checks if the the group is overlapping another group or sprite.
   * The check is performed using the colliders. If colliders are not set
   * they will be created automatically from the image/animation bounding box.
   *
   * A callback function can be specified to perform additional operations
-  * when the overlap occours.
+  * when the overlap occurs.
   * The function will be called for each single sprite overlapping.
   * The parameter of the function are respectively the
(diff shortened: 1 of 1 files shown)


lib/p5.play.js
```diff
@@ -2712,7 +2712,7 @@ function Group() {
   * </code>
   *
   * @method overlap
-  * @param {Object} target Group to check against the current one
+  * @param {Object} target Group or Sprite to check against the current one
```
islemaster: You could be more specific in the parameter type here (pretty sure; YUIDoc reference for [@type](http://yui.github.io/yuidoc/syntax/index.html#type), [@param](http://yui.github.io/yuidoc/syntax/index.html#param)):

```
* @param {Group|Sprite} Group or Sprite to check against the current one
```

islemaster:
Once comment on documenting types, otherwise looks good to me!

islemaster:
Want to fix #63 while you're at it here?

lib/p5.play.js
toolness: Oh, I had considered that, but I figured there were other places where `{Object}` was used in place of something more specific, so I just changed the wording... Do you mind if we file a separate PR for that so I can change all the `{Object}` types in one go?

toolness:
> Want to fix #63 while you're at it here?

Actually do you mind if I do that in a separate PR? I wanted to keep this one only focused on the docs for `Group`... maybe it's a bit pedantic but I guess I prefer a bunch of tiny focused PRs to bigger ones. Granted, I guess this one isn't particularly focused to begin with. :)

toolness:
Uh, also, it turns out that figuring out exactly *how* we want to fix #63 is actually a bit non-trivial, so I've filed https://github.com/molleindustria/p5.play/pull/67 separately. :grin:

islemaster:
If you want to break stuff out, I'm all for it.  This LGTM!

toolness:
Cool thanks!

PR #70, opened 2016-03-30
Clean lint and run linter before tests

### LINKED ISSUE #60: Fix lint errors and run linter as part of test suite
Alternatively, we could just have the linter detect for undeclared globals, but I'm not sure if that's possible with jshint... My main intent here is to prevent regressions like #59.

islemaster:
Fixes #60.

Fixes lint errors and sets up `npm run lint` as a pretest step.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -3336,13 +3336,14 @@ function Animation(pInst) {
       //same digits: case img0001, img0002
       else
       {
+        var fileName;
         if(digits1 === digits2)
         {
 
           //load all images
           for (var i = number1; i <= number2; i++) {
             // Use nf() to number format 'i' into four digits
-            var fileName = prefix1 + pInst.nf(i, digits1) + ".png";
+            fileName = prefix1 + pInst.nf(i, digits1) + ".png";
             this.images.push(pInst.loadImage(fileName));
 
           }
@@ -3353,7 +3354,7 @@ function Animation(pInst) {
           //print("from "+prefix1+" "+number1 +" to "+number2);
           for (var i = number1; i <= number2; i++) {
             // Use nf() to number format 'i' into four digits
-            var fileName = prefix1 + i + ".png";
+            fileName = prefix1 + i + ".png";
             this.images.push(pInst.loadImage(fileName));
 
           }
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -181,9 +181,9 @@ p5.prototype.removeSprite = function(sprite) {
 */
 p5.prototype.updateSprites = function(upd) {
 
-  if(upd==false)
+  if(upd === false)
     this.spriteUpdate = false;
-  if(upd==true)
+  if(upd === true)
     this.spriteUpdate = true;
 
   if(this.spriteUpdate)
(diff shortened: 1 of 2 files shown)


lib/p5.play.js
```diff
@@ -3199,7 +3197,8 @@ defineLazyP5Property('AABB', boundConstructorFactory(AABB));
    */
 
 function Animation(pInst) {
-  arguments = Array.prototype.slice.call(arguments, 1);
+  var frameArguments = Array.prototype.slice.call(arguments, 1);
```
toolness: Ah, thanks for this, I had felt queasy about reassigning to `arguments` when I wrote this, and was too lazy at the time to do what you just did :)

lib/p5.play.js
```diff
@@ -250,7 +250,7 @@ p5.prototype.drawSprites = function(group) {
   }
   else if(arguments.length===1)
   {
-    if(arguments[0] instanceof Array == false)
+    if(arguments[0] instanceof Array === false)
```
toolness: Not that this is really relevant to this PR specifically, but I'm wondering if we should eventually use `Array.isArray()` here to avoid the [`instanceof` and multiple contexts/windows problem](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_context_%28e.g._frames_or_windows%29), and just adhere to best practices in general? I could file an issue for it if you agree, or whatevs we can just let it go.

toolness:
Yeah, this is awesome!! Thanks!

lib/p5.play.js
islemaster: Yes, I have a set of other cleanup changes on the way too. This was just
the 'minimum' to pass our lint configuration.

GoToLoop: Strictly checking whether an object is a "real" _**Array**_ leaves out typed arrays and other pseudo array-like structures like _arguments_:  

1. https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray

2. https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/arguments

Instead of `instanceof Array` or **isArray()** über strict checks, something like this is pretty enough to identify all types of array-like objects: `typeof obj === 'object' && obj.length >= 0`  :-D

islemaster: While that flexibility is nice, here we are explicitly trying to validate that the argument is a `Group`, and don't want to support other array-like objects.

In a perfect world we'd ask `arguments[0] instanceof Group`, but since `Group` is a sort of augmented array instead of being defined by prototypal inheritance it won't work.

On the other hand, if we wanted to use duck-typing in this case the proper approach would probably be to accept any argument with a `draw()` method.  That's probably a better idea than checking for any sort of array... as it is, you could slip a plain array past this check and throw on line 257.  I'll make that change in another PR.

toolness: I think by "another PR" @islemaster means https://github.com/molleindustria/p5.play/pull/77... Not sure why GitHub didn't auto-mention it here.

PR #72, opened 2016-03-30
Fix errors in Sprite docs.

toolness:
This fixes some inconsistencies in the Sprite docs.

It also gets rid of the following spurious method entry:


islemaster:
:+1: Nice cleanup.

PR #74, opened 2016-03-30
Simplify argument handling

islemaster:
We were handling function arguments in a needlessly complex way in a few cases.  Explanations inline.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -234,30 +234,15 @@ p5.prototype.getSprites = function() {
 * @param {Group} [group] Group of Sprites to be displayed
 */
 p5.prototype.drawSprites = function(group) {
+  // If no group is provided, draw the allSprites group.
+  group = group || this.allSprites;
 
-  //draw everything
-  if(arguments.length===0)
+  if (!(group instanceof Array))
   {
-    //sort by depth
-    this.allSprites.sort(function(a,b) {
-      return a.depth - b.depth;
-    });
-
-    for(var i = 0; i<this.allSprites.size(); i++)
-    {
-      this.allSprites.get(i).display();
-    }
+    throw("Error: with drawSprites you can only draw all sprites or a group");
   }
-  else if(arguments.length===1)
-  {
-    if(arguments[0] instanceof Array === false)
-      throw("Error: with drawSprites you can only draw all sprites or a group");
-    else
-    {
-      arguments[0].draw();
-    }
 
-  }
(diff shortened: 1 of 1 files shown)


lib/p5.play.js
```diff
@@ -234,30 +234,15 @@ p5.prototype.getSprites = function() {
...
-  {
-    if(arguments[0] instanceof Array === false)
-      throw("Error: with drawSprites you can only draw all sprites or a group");
-    else
-    {
-      arguments[0].draw();
-    }
 
-  }
+  group.draw();
```
islemaster: I carefully compared the implementations of `group.draw()` and the sort/display work being done on `allSprites` here, and discovered that they're identical.  Therefore, the whole implementation of this method collapses into a few lines.

lib/p5.play.js
```diff
@@ -1820,8 +1805,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * @return {Boolean} result True if non-transparent
   */
   this.overlapPixel = function(pointX, pointY) {
-
-    var point = createVector(arguments[0], arguments[1]);
+    var point = createVector(pointX, pointY);
```
islemaster: There was no reason to get the named arguments `pointX` and `pointY` by indexing into `arguments` here.

lib/p5.play.js
```diff
@@ -1855,7 +1839,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * @return {Boolean} result True if inside
   */
   this.overlapPoint = function(pointX, pointY) {
-    var point = createVector(arguments[0], arguments[1]);
+    var point = createVector(pointX, pointY);
```
islemaster: Likewise re: named arguments here.

lib/p5.play.js
```diff
@@ -3413,21 +3397,18 @@ function Animation(pInst) {
...
+   *
+   * @method draw
+   * @param {Number} x x coordinate
+   * @param {Number} y y coordinate
+   * @param {Number} [r=0] rotation
+   */
+  this.draw = function (x, y, r) {
+    this.xpos = x;
+    this.ypos = y;
+    this.rotation = r || 0;
```
islemaster: `Animation.draw()` had an undocumented optional third argument.  I've documented it, simplified its use, and replaced the unnecessary uses of `arguments` with named parameters.

toolness:
Nice, this is great! Thanks.

PR #76, opened 2016-03-30
Adopt ESLint instead of JSHint

islemaster:
I'm going to go out on a limb here and suggest that we move to ESLint.  I'm sure there are lots of arguments on both sides of this one, but here are my favorite reasons:

1. [JSHint is trying to migrate away from enforcing code style.](http://jshint.com/blog/jshint-3-plans/)  Deprecation warnings when you run it suggest that their goal is to catch potential errors, not enforce style.
1. [ESLint can perform automatic fixes for many of its options.](http://eslint.org/docs/user-guide/command-line-interface#fix)

It's gained momentum recently; [Dan Abramov says it's smarter than JSHint](https://medium.com/@dan_abramov/lint-like-it-s-2015-6987d44c5b48#.rsys4ci2w) and it gets called out in this [State of the Art JavaScript in 2016](https://medium.com/javascript-and-opinions/state-of-the-art-javascript-in-2016-ab67fc68eb0b#3cd7) article.

A few notes about this implementation:

## .eshintrc.js is huge
Don't worry, I'm not adding tons of rules here.  That config file was actually generated by running `eshint --init` and having it examine p5.play.js and make its best guess at our current style guidelines.  Then I made some manual adjustments and (see the commit list) enabled some additional rules that were easy to support (often thanks to the autocorrect feature).

## Automatic fixes don't happen by default
`npm run lint` (and by extension, `npm run test`) are going to complain about things ESLint knows how to fix automatically - for now.  Mostly this is because don't I want our CircleCI run ignoring linting errors because they get auto-fixed locally.  It would be nice to set up a precommit hook that did linting with fixes.

In the meantime, run `npm run lint -- --fix` to have ESLint clean a bunch of stuff up for you.

## We're only linting p5.play.js
...which is what JSHint was doing too.  I've been doing local prototyping of linting for our unit tests and for the examples.  I'll submit those in separate PRs.


.eslintrc.js
```diff
@@ -0,0 +1,225 @@
...
+      "last"
+    ],
+    "complexity": "off",
+    "computed-property-spacing": "error",
+    "consistent-return": "off",
+    "consistent-this": "off",
+
+    // Use of braces is very inconsistent in the library so far.  It will
+    // be a significant change to enable this check.
+    "curly": "off",
```
toolness: Yeah, good choice! I really miss comments in JSON. And am constantly shooting myself in the foot with YML. :grin:

toolness:
This looks great! Thanks for all the notes on the rationale behind these changes, too.

.jshintrc
```diff
@@ -1,12 +0,0 @@
-{
-  "eqnull": true,
```
GoToLoop: Both `== null` & `== undefined` catches both `undefined` & `null` in 1 swoop.  
It's true it's rare in JS to have variables w/ `null` values.  
Nonetheless it's highly recommended allowing both `== null` & `== undefined` in order to make sure both `undefined` & `null` are caught.

islemaster: It's a personal preference to always treat them differently.  I'll submit a change to re-relax this rule.

Something I forgot to consider in this change is the value of consistency with the [p5.js jshint](https://github.com/processing/p5.js/blob/master/.jshintrc) and [jscs](https://github.com/processing/p5.js/blob/master/.jscsrc) configs.  I'll do another PR that updates our style rules for that.

PR #77, opened 2016-03-30
Duck-type group before drawing.

islemaster:
[This discussion](https://github.com/molleindustria/p5.play/pull/70/files#r57978423) led me to the conclusion that it would be less error-prone to check for the `draw()` method on the argument to `drawSprites()` than to check that it is an array - not all arrays can be `draw()`n anyway.


toolness:
Oh, interesting. Makes sense to me, thanks!

Issue #43, opened 2016-02-28
mouseUp, mouseDown, keyDown are inaccurate for 1 frame


toolness:
The folowing was originally noted in #42.

The documentation for `keyDown` states:

```
Like p5 keyIsDown but accepts strings and codes
```

However, the implementation is checking to see if the internal key state is specifically equal to `KEY_IS_DOWN`. The problem is that the internal keystate might actually be `KEY_WENT_DOWN`, in which case the method will retrun `false` but `keyIsDown()` will be `true`!

It's understandable that no one has noticed this, because it just means that the result of `keyDown` will be inconsistent with `keyIsDown` for only the frame in which the key went from up to down.

A parallel problem exists with `mouseDown`/`mouseIsDown`, as well as `mouseUp`.

@islemaster has some [thoughts](https://github.com/molleindustria/p5.play/pull/42#issuecomment-189685433) on it:

> I knew about this. I definitely want to fix it so keyDown and keyWentDown
> are not mutually exclusive states. I figured it wouldn't happen soon
> though, since it's technically a breaking change. I also want to check
> whether @molleindustria did this on purpose, to circumvent a certain set of
> beginner mistakes.
> 
> If we had a release process and semver I would bundle this with other
> breaking changes in a major version. Since things are still pretty fast and
> loose (and I don't think anyone auto-updates this library?) maybe we can
> just fix it.

quinton-ashley:
@toolness @islemaster Is this still a problem?

islemaster:
Hi @quinton-ashley! Thanks for maintaining here.  Yes, this was never addressed - the `keyDown` method still only checks for the `KEY_IS_DOWN` state, and its documentation hasn't changed.

https://github.com/molleindustria/p5.play/blob/b6ffefd86a9920e13f5248bb05da22df27332a5b/lib/p5.play.js#L3759-L3769

That said, I'm not sure it's a priority. 🤷 If we have no user complaints about this behavior it might be better to leave it.

quinton-ashley:
Thanks! I will change this in v3. I made a wrapper for the p5.js keyIsDown function that converts key names to key codes, instead of reimplementing it.

quinton-ashley:
wontfix for v2, closing this issue

PR #79, opened 2016-03-31
Lint tests

islemaster:
Add linting to our test files, with a few special rules defined in test/unit/.eslintrc.js (and documented inline there).  This was super-easy, because our tests were mostly clean already!

### DIFF
--- a/test/unit/.eslintrc.js
+++ b/test/unit/.eslintrc.js
@@ -8,6 +8,9 @@ module.exports = {
     "sinon": true
   },
   "rules": {
+    // Eventually move up to root config.
+    "indent": ["error", 2],
+
     // Eventually we want to propagate this setting up to the main config.
     "keyword-spacing": "error",
 
--- a/test/unit/.eslintrc.js
+++ b/test/unit/.eslintrc.js
@@ -1,0 +1,20 @@
+module.exports = {
+  "env": {
+    "mocha": true
+  },
+  "globals": {
+    "expect": true,
+    "p5": true
+  },
+  "rules": {
+
+    // Allow empty functions in tests, because we often want to pass a no-op
+    // placeholder when testing.
+    // Very frequent: new p5(function () {})
+    "no-empty-function": "off",
+
+    // Chai BDD assertions violate this style rule, so disable it in tests
+    // e.g. expect(value).to.equal('something')
+    "no-unused-expressions": "off"
+  }
+};
--- a/test/unit/.eslintrc.js
+++ b/test/unit/.eslintrc.js
@@ -8,6 +8,9 @@ module.exports = {
     "sinon": true
   },
   "rules": {
+    // Eventually we want to propagate this setting up to the main config.
+    "keyword-spacing": "error",
+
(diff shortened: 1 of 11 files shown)


test/unit/keycodes.js
```diff
@@ -17,81 +17,94 @@ describe('_keyCodeFromAlias', function() {
...
-            return char.toUpperCase();
-          } else {
-            return char.toLowerCase();
-          }
-        }).join('');
-
-        expect(_keyCodeFromAlias(upperCaseAlias))
-          .to.equal(_keyCodeFromAlias(lowerCaseAlias))
-          .to.equal(_keyCodeFromAlias(randomCaseAlias));
+        if (KEY.hasOwnProperty(alias)) {
```
islemaster: The change here is the `hasOwnProperty` check right inside the `for(...in...)` loop.  There are a few of these.

islemaster:
Note: This PR introduces warnings for TODOs, and there are three in tests that I'm not fixing (visible in the Travis log):
```
/home/travis/build/molleindustria/p5.play/test/unit/input.js
   46:5  warning  Unexpected 'todo' comment  no-warning-comments
   64:5  warning  Unexpected 'todo' comment  no-warning-comments
  110:5  warning  Unexpected 'todo' comment  no-warning-comments
✖ 3 problems (0 errors, 3 warnings)
```

[These TODOs are references](https://github.com/molleindustria/p5.play/blob/master/test/unit/input.js#L46) to issue #43 about key handlers potentially being inaccurate for one frame, and should be removed when that issue is resolved.

test/unit/.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+    // Very frequent: new p5(function () {})
+    "no-empty-function": "off",
+
+    // Chai BDD assertions violate this style rule, so disable it in tests
+    // e.g. expect(value).to.equal('something')
+    "no-unused-expressions": "off",
+
+    // Eventually we want to propagate these four settings up to the main config.
+    "space-before-blocks": "error",
+    "space-before-function-paren": "error",
```
toolness: WAT?!! But `function foo ()` looks so weird to me... :stuck_out_tongue_winking_eye: 

Ah, it's all good, I'll get over it.

toolness:
Looks good to me, thanks!

test/unit/.eslintrc.js
islemaster: Oh, I'm not sure this is what I meant it to me.  I propose updating this to use the setting [`{anonymous: 'always', named: 'never'}`](http://eslint.org/docs/rules/space-before-function-paren#anonymous-always-named-never).

```javascript
// Required
function foo() {};
var foo = function () {};

// Broken
function foo () {};
var foo = function() {};
```

Rationale being if you put parens adjacent to a keyword (`function()`) you should do it everywhere (`if()`, `for()`). Objections?

islemaster: Never mind - we _almost_ follow the "never" version of this rule in p5.play.js, I'm going to switch to that.

PR #80, opened 2016-03-31
Remove built docs from repository

### LINKED ISSUE #57: Remove built docs from repository
It would be really nice if we could remove the built docs from the repository, so that we don't need to spam the commit history with rebuilds of the documentation whenever we change it. 

(This issue was split off from #41.)

toolness:
This is an attempt to fix #57.

It removes the `docs` directory from the development branch, adds an `update-gh-pages.sh` script that rebuilds the docs and copies them, the `examples` directory, and `lib/p5.play.js` to the `gh-pages` branch and stages the changes.

I've already used it on the previously-nonexistent `gh-pages` branch, which you can see here:

http://molleindustria.github.io/p5.play/

If this looks good, after merging we just need to:

1. Add a `CNAME` file to the root of the `gh-pages` branch containing the text `p5play.molleindustria.org`.
2. Add a DNS CNAME record for **p5play.molleindustria.org** to point at **molleindustria.github.io**.

Then we're good to go!


toolness:
Er, sorry the diff looks like hell. It's probably easier to just look at all the commits except the one where I removed the `docs` directory:

* https://github.com/molleindustria/p5.play/pull/80/commits/271e22f92fbaf467b7a4cb7198faea4aff7fc98c Add `update-gh-pages.sh`
* https://github.com/molleindustria/p5.play/pull/80/commits/61c3575d186c90148be700c4a3ff0851982856fe Tell users to generate docs after npm install.
* https://github.com/molleindustria/p5.play/pull/80/commits/8b431bf5e1b3cc0357c6ece3d489b1d56f2c88b9 add 'docs' to .gitignore.

islemaster:
We should check out [Auto-deploying built products to gh-pages with Travis](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd) too; looks like with a little more config, we can keep the docs updated automatically.

update-gh-pages.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/sh
+
+set -eu
+
+echo "Regenerating documentation."
+
+rm -rf docs
+npm run docs
```
islemaster: I _think_ yuidoc implicitly wipes out the docs directory.  It doesn't do proper incremental updates in any case.  I think we should just make `rm -rf docs` part of the `npm run docs` task.

README.md
```diff
@@ -19,6 +19,7 @@ First install [node.js][]. Then run:
 
 ```
 npm install
+npm run docs
```
islemaster: Could just add `npm run docs` as a postinstall step.

.gitignore
```diff
@@ -1,2 +1,3 @@
 .idea
 node_modules
+docs
```
islemaster: Do we also need to add `__temp_examples`, `__temp_docs` and `__temp_p5.play.js` just in case?

islemaster:
On the whole, this is pretty slick.  I like it, and I like the prospect of keeping the hosted docs up to date.  We can do the automatic rebuild as a separate step; that doesn't need to block this change.

update-gh-pages.sh
toolness: Hmm, so while I generally agree with that, the only thing is that, um, `rm` isn't actually available on Windows, so doing that would break things for Windows contributors who don't have cygwin or Git bash installed. :disappointed: 

I settled for making `update-gh-pages.sh` a shell script because updating the website only really needs to be done by maintainers and Travis, but if just building the docs is broken for Windows users by default, that sort of sucks... Mind if we postpone this for now, and maybe bring in a more cross-platform solution later?  Or we could wait until [Bash comes to Windows](https://insights.ubuntu.com/2016/03/30/ubuntu-on-windows-the-ubuntu-userspace-for-windows-developers/) and then do whatever we want! :stuck_out_tongue_winking_eye:

toolness:
Awesome, thanks! I made all the changes you suggested, with the exception of `rm -rf docs` as part of `npm run docs`, which I explained [over here](https://github.com/molleindustria/p5.play/pull/80#discussion_r58147665).

update-gh-pages.sh
islemaster: :facepalm: I've been in *nix-land too long.  Nice catch.

islemaster:
Looks like there's a lint failure on Travis.

toolness:
Yeah, I'm confused by it because it's reporting this:

```
/home/travis/build/molleindustria/p5.play/examples/player.js
  8:18  error  Unexpected space before function parentheses  space-before-function-paren
```

But that file wasn't changed in my PR...

islemaster:
:+1: Was the result of merging my other two PRs at about the same time.  This looks good!

toolness:
Cool, thanks!

PR #81, opened 2016-03-31
Lint examples

islemaster:
Lint JavaScript in the examples directory (primarily the examples themselves).  Uses custom, significantly relaxed rules (defined and documented in examples/.eslintrc.js) since examples assume a lot about the environment they're run in.


toolness:
Nice, thanks for the helpful comments explaining the rationale for all the linting options!

PR #82, opened 2016-03-31
Change space-before-function-paren setting.

islemaster:
Require no space between function name and parens for both named and anonymous functions (see [discussion here](https://github.com/molleindustria/p5.play/pull/79/files/639d24dec5dae7f1ab869027627d3b63aec242e2#r58138388)).

Move setting up to root config, since we nearly conform in p5.play.js already.

Autocorrect style.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -197,11 +197,19 @@ module.exports = {
     "sort-imports": "error",
     "sort-vars": "off",
 
-    // I'd like to enable these four rules, but should discuss target style
-    // with the community before we do because we don't follow these at all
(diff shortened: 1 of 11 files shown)


islemaster:
You prefer this?

toolness:
Oh, heh, I do prefer it, though the rationale your provided in https://github.com/molleindustria/p5.play/pull/79#discussion_r58143199 for the current behavior actually makes sense too, so I'm OK with it either way. I guess since you're asking, though, I'll happily switch back to the style I'm used to--we can always easily switch back later if we want anyhow, thanks to ESLint's auto-correcting!

PR #85, opened 2016-04-07
lock rotateToDirection bidirectionally, make angle optional on setSpeed

cpirich:
code.org is using p5.play.js as part of a new computer science curriculum. We would like to change the behavior of the Sprite object's `rotateToDirection` property to make it lock the `rotation` property with the `velocity` angle bi-directionally (the current behavior is to ignore the `rotation` property in this mode and continually overwrite it with the angle of the `velocity` vector). This change is accomplished by implementing a custom "setter" for the `rotation` property that updates the `velocity` vector when `rotateToDirection` is `true`. The internal (private) storage for the `rotation` property value is now stored in the `rotation_` property.

In addition, we would like to change the second parameter of the Sprite object's `setSpeed` function (the `angle`) to be optional, so that the speed alone can be changed. (Omitting the `angle` parameter is essentially the same as calling `sprite.setSpeed(<new value>, sprite.getDirection())`, except we have elected to default the `angle` to the `rotation` property when the `velocity` vector is `0,0`). This allows the student to enable `rotateToDirection`, then freely call `setSpeed()` with a `speed` parameter only and set `rotation` to change direction. The sprite can be stopped in this mode by calling `setSpeed(0)` and restarted along the same angle by calling `setSpeed()` with a positive speed value.

The `rotationSpeed` property will now continue to be respected even while in `rotateToDirection` mode. To preserve the previous behavior of always locking the `rotation` property to the `velocity` angle, the `rotation` property should not be modified and the `rotationSpeed` property should be set to zero.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -892,7 +892,29 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * @type {Number}
   * @default 0
   */
-  this.rotation = 0;
+  Object.defineProperty(this, 'rotation', {
+    enumerable: true,
+    get: function() {
+      return this.rotation_;
+    },
+    set: function(value) {
+      this.rotation_ = value;
+      if (this.rotateToDirection) {
+        this.setSpeed(this.getSpeed(), value);
+      }
+    }
+  });
+
+  /**
+  * Internal rotation variable (expressed in degrees).
+  * Note: external callers access this through the rotation property above.
+  *
+  * @private
+  * @property rotation_
+  * @type {Number}
+  * @default 0
+  */
+  this.rotation_ = 0;
 
   /**
   * Rotation change in degrees per frame of thevisual element (image or animation)
@@ -906,8 +928,8 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
 
   /**
-  * Automatically set the rotation of the visual element
-  * (image or animation) to the sprite's movement direction.
+  * Automatically lock the rotation property of the visual element
+  * (image or animation) to the sprite's movement direction and vice versa.
   *
   * @property rotateToDirection
   * @type {Boolean}
@@ -1115,10 +1137,10 @@ function Sprite(pInst, _x, _y, _w, _h) {
       if(this.maxSpeed !== -1)
         this.limitSpeed(this.maxSpeed);
 
-      if(this.rotateToDirection)
-        this.rotation = this.getDirection();
-      else
-        this.rotation += this.rotationSpeed;
+      if(this.rotateToDirection && this.velocity.mag() > 0)
+        this.rotation_ = this.getDirection();
+
+      this.rotation += this.rotationSpeed;
 
       this.position.x += this.velocity.x;
       this.position.y += this.velocity.y;
@@ -1616,13 +1638,25 @@ function Sprite(pInst, _x, _y, _w, _h) {
   /**
   * Set the speed and direction of the sprite.
   * The action overwrites the current velocity.
+  * If direction is not supplied, the current direction is maintained.
+  * If direction is not supplied and there is no current velocity, the current
+  * rotation angle used for the direction.
   *
   * @method setSpeed
-  * @param {Number}  speed Scalar speed to add
-  * @param {Number}  angle Direction in degrees
+  * @param {Number}  speed Scalar speed
+  * @param {Number}  [angle] Direction in degrees
   */
   this.setSpeed = function(speed, angle) {
-    var a = radians(angle);
+    var a;
+    if (typeof angle === 'undefined') {
+      if (this.velocity.x !== 0 || this.velocity.y !== 0) {
+        a = Math.atan2(this.velocity.y, this.velocity.x);
+      } else {
+        a = radians(this.rotation_);
+      }
+    } else {
+      a = radians(angle);
+    }
     this.velocity.x = cos(a)*speed;
     this.velocity.y = sin(a)*speed;
   };
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -895,10 +895,10 @@ function Sprite(pInst, _x, _y, _w, _h) {
   Object.defineProperty(this, 'rotation', {
     enumerable: true,
     get: function() {
-      return this.rotation_;
+      return this._rotation;
     },
     set: function(value) {
-      this.rotation_ = value;
+      this._rotation = value;
       if (this.rotateToDirection) {
         this.setSpeed(this.getSpeed(), value);
       }
@@ -910,11 +910,11 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * Note: external callers access this through the rotation property above.
   *
   * @private
-  * @property rotation_
+  * @property _rotation
   * @type {Number}
   * @default 0
   */
-  this.rotation_ = 0;
+  this._rotation = 0;
 
   /**
   * Rotation change in degrees per frame of thevisual element (image or animation)
@@ -1138,7 +1138,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
         this.limitSpeed(this.maxSpeed);
 
       if(this.rotateToDirection && this.velocity.mag() > 0)
-        this.rotation_ = this.getDirection();
+        this._rotation = this.getDirection();
 
       this.rotation += this.rotationSpeed;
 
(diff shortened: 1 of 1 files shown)


cpirich:
@islemaster @toolness

lib/p5.play.js
```diff
@@ -892,7 +892,29 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
+  /**
+  * Internal rotation variable (expressed in degrees).
+  * Note: external callers access this through the rotation property above.
+  *
+  * @private
+  * @property rotation_
+  * @type {Number}
+  * @default 0
+  */
+  this.rotation_ = 0;
```
islemaster: p5.play's convention for private members is a leading underscore, not a trailing once.  Please rename to `_rotation`.

lib/p5.play.js
```diff
@@ -1616,13 +1638,25 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
+  * @param {Number}  [angle] Direction in degrees
   */
   this.setSpeed = function(speed, angle) {
-    var a = radians(angle);
+    var a;
+    if (typeof angle === 'undefined') {
+      if (this.velocity.x !== 0 || this.velocity.y !== 0) {
+        a = Math.atan2(this.velocity.y, this.velocity.x);
+      } else {
+        a = radians(this.rotation_);
```
islemaster: I'm not sure how I feel about this interaction.  It produces some funny-ish edge cases, like this one:

1. Given `rotateToDirection` is false
2. And my sprite rotation is 0 and the sprite is moving at a 45° angle at a speed of 1
3. When I call `setSpeed(0.5)`
4. Then my sprite rotation is 0 and the the sprite is moving at a 45° angle at a speed of 0.5
5. When I call `setSpeed(0)`
6. Then my sprite rotation is 0 and the the sprite is not moving
7. When I call `setSpeed(1)`
8. Then my sprite rotation is 0 and the the sprite is moving at a 0° angle at a speed of 1

It seems odd that in some cases `setSpeed` changes the sprite's direction of motion and in other cases it doesn't.

lib/p5.play.js
```diff
@@ -1616,13 +1638,25 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
-  * @param {Number}  angle Direction in degrees
+  * @param {Number}  speed Scalar speed
+  * @param {Number}  [angle] Direction in degrees
   */
   this.setSpeed = function(speed, angle) {
-    var a = radians(angle);
+    var a;
+    if (typeof angle === 'undefined') {
+      if (this.velocity.x !== 0 || this.velocity.y !== 0) {
+        a = Math.atan2(this.velocity.y, this.velocity.x);
```
islemaster: Why don't we just call `this.getDirection()` here?

islemaster:
Chris, I see this as three distinct (if somewhat entangled) changes:

1. `rotateToDirection` locks rotation and movement bidirectionally.
2. The second argument to `setSpeed` is optional
3. `rotationSpeed` is not ignored when `rotateToDirection` is true.

I think 2 and 3 are somewhat dependent on 1, at least to behave like we think they should.  Is there a reason not to do 1 in isolation first?

We discussed 1 and 2 in person, but 3 was a surprise to me.  It's kind of cool that you can make a sprite move in a circle with

```javascript
sprite.rotateToDirection = true;
sprite.rotationSpeed = 1;
sprite.setSpeed(2);
```

1. Since rotation is previously ignored when `rotateToDirection` is true, giving it a special behavior when `rotateToDirection` is true just introduces new behavior.  Only possible break is if someone was somehow depending on it being a no-op.
2. Just makes an argument optional that wasn't optional before, so shouldn't break anything.
3. Given `rotation` was ignored when `rotateToDirection` was true, so was `rotationSpeed` which only acted on `rotation` - so again, only breaking if somebody depended on the no-op.

lib/p5.play.js
cpirich: `setSpeed` should never change the direction of motion. It does, however, need to pick a direction of motion if there is no velocity, since the "last direction of motion" is not preserved and that would be a more complex change. In the `rotateToDirection` case, this allows the `rotation` property to be used to preserve the last direction of motion. Since this is new behavior (calling it without the 2nd parameter), it won't break anything and it is in YUIDoc comments.

FYI, the previous equivalent (attempting to change speed only by using `sprite.setSpeed(<new speed>, sprite.getDirection())` would have the same issue with the scenario you described above because the heading angle is lost once the velocity becomes `0, 0`

lib/p5.play.js
cpirich: We could do that, but we need it in radians, so getDirection() would convert to degrees and we'd turn around and convert it back to radians.

cpirich:
Agree with your summary. We can separate these into different pull requests, but if all of the changes are acceptable, it is easier to do them together since the behavior is definitely related. (e.g. it doesn't make sense to start paying attention to `rotationSpeed` while `rotateToDirection` is `true` without these other changes).

The decision to change the behavior of `rotationSpeed` while in `rotateToDirection` mode came as a result of a discussion with @poorvasingal during code inspection. Since the no-op provided no value and the old behavior can be preserved by setting `rotationSpeed` to zero, it seems consistent if we're going to honor `rotation` in this mode, we should also honor `rotationSpeed`.

And, yes, I like the fact that it enables you to do some very simple movement using code like the example you created above.

lib/p5.play.js
cpirich: Good point, will fix.

lib/p5.play.js
islemaster: Although `sprite.setSpeed(_, sprite.getDirection())` exhibits the same problem, I feel like the problem is less hidden, given that `getDirection()` calls out its own inability to give a reasonable answer when velocity is 0.

I'd think I'd prefer it if the single-argument version of `setSpeed()` instead behaved like `sprite.setSpeed(_, sprite.rotation)`.  It provides the same behavior when `rotateToDirection` is true, and (IMO) a more consistent, less surprising behavior when it's false.

(I understand our curriculum team might disagree though.  Have we discussed this case with them?  @poorvasingal do you know?)

mrjoshida: So if that's the case would rotateToDirection really only impact the visible rotation of the sprite? Meaning that rotation is always correlated with direction?

islemaster: Quick in-person talk with Poorva led to a use case for one-argument `setSpeed()`.  Suppose you did the following right when your sketch started:

```javascript
sprite.pointTo(10, 10);
sprite.setSpeed(2);
```

The expected behavior here is to face (10, 10) and start moving toward it, right?   I'll admit that your current implementation does this, but using `rotation` as the default value would do it too, with fewer edge cases.

Here's a clumsy attempt to illustrate the relative complexities of the approaches, by illustrating the effective default value for the second argument in different situations

Sprite state | Current impl | `rotation` default impl
---|---|---
`rotateToDirection` and `velocity`==0 | Use `rotation` | Use `rotation`
`rotateToDirection` and `velocity`!=0 | Use `getDirection()` (which is the same as `rotation`) | Use `rotation`
`!rotateToDirection` and `velocity`==0 | Use `rotation` | Use `rotation`
`!rotateToDirection` and `velocity`!=0 | Use `getDirection()` | Use `rotation`

The current implementation _effectively_ uses `rotation` in 3 out of 4 cases already.

islemaster: @mrjoshida My concern here is with the `rotateToDirection=false` case.  I'm not proposing a change to the (proposed) `rotateToDirection=true` behavior.

cpirich: Interesting, because I think it is very odd for the `rotation` property to have any impact on `setSpeed()` when `rotateToDirection` is `false`. I think if a sprite is in motion, it will be rare for it to be rotated (except for top-down games). So if the caller simply wants to change the speed (not the direction), they could call `setSpeed()` with just one scalar parameter.

poorvasingal: @cpirich - That's what I was thinking earlier too but then @islemaster repeated your concern about the behavior being different based on the speed (e.g. speed being zero vs. really slow, or just non-zero in general).  I think it makes sense to try to have consistent behavior regardless of what the speed is.  The other thing that Brad brought up was that if you're rotated in a certain direction and want to set your speed for the first time to a different direction (e.g. just going up for example), you're going to need to call setSpeed with two parameters regardless.  So even if you're trying to increase your speed, you would already need to know what direction you were traveling in.

cpirich: If we want consistent behavior, I'd argue that we should always ignore the rotation property when rotateToDirection is false. Direction of movement is independent of rotation in that mode.

poorvasingal: @cpirich I brought up the complexities with the curriculum team and it now sounds like they are leaning towards just sticking to a 2 parameter setSpeed method.  So we might be able to avoid this whole conversation entirely.  I'm still waiting for a final confirmation, but we should definitely hold off on any changes related to setSpeed.

cpirich:
Closing - will re-open with merges re-applied

PR #87, opened 2016-04-29
Document sprite.depth default value

islemaster:
_Fixes #75_

Expands documentation for sprite.depth.

While I was at it, I couldn't resist cleaning up the implementations of `Group.maxDepth` and `Group.minDepth`, which meant writing corresponding tests to pin the existing behavior.  These tests reveal a few weird edge cases around existing behavior, but I'm not changing it in this PR, just pinning it.  We can change later if we want.

Weird behaviors:
* Default value of `maxDepth` is 0 when a group is empty, but a group can contain sprites with only negative depths which will give a `maxDepth` less than 0.  A better default might be `-Infinity`.
* Stranger: Default value of `minDepth` is 99,999 when a group is empty, but a group can contain sprites with only depths greater than 99,999 in which case `minDepth` is also greater.  A better default might be `Infinity`.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -139,6 +139,10 @@ p5.prototype.spriteUpdate = true;
    * A Sprite can have a collider that defines the active area to detect
    * collisions or overlappings with other sprites and mouse interactions.
    *
+   * Sprites created using createSprite (the preferred way) are added to the
+   * allSprites group and given a depth value that puts it in front of all
+   * other sprites.
+   *
    * @method createSprite
    * @param {Number} x Initial x coordinate
    * @param {Number} y Initial y coordinate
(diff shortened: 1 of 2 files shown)


lib/p5.play.js
```diff
@@ -2580,18 +2586,13 @@ function Group() {
...
-
-    for(var i = 0; i<array.length; i++)
-      if(array[i].depth>max)
-        max = array[i].depth;
+    if (array.length === 0) {
+      return 0;
+    }
 
-    return max;
+    return array.reduce(function(memo, sprite) {
```
toolness: That's interesting that you've called the first argument `memo`... is that a convention? I'd have thought that naming it something like `maxDepth` might be more readable, but this is fine too.

toolness:
Cool thanks! I agree with your ideas about the better defaults for the weird behaviors, too.

Feel free to rename the `memo` variables if you think it might make things more readable, or if it's a functional programming naming convention I'm just unaware of or something like that, it's cool to keep it.

Everything else looks great--merge at will!

lib/p5.play.js
islemaster: Good catch, I totally didn't think about it.  `memo` is a convention I've seen for the first argument to a `reduce` callback (see [the Underscore.js documentation](http://underscorejs.org/#reduce)) but I forget that it's not widespread, and you're 100% right that `maxDepth` reads better here.

PR #86, opened 2016-04-29
Update p5.js version in examples/lib to 0.4.24

islemaster:
Previously at 0.4.21. No other changes to code obviously required; tests still pass.

0.4.24 includes [this bugfix](https://github.com/processing/p5.js/pull/1321) which should resolve issue #54 with Sprite.debug drawing.

See 0.4.24 release notes here: https://github.com/processing/p5.js/releases/tag/0.4.24

PR #88, opened 2016-05-04
Revert "Update p5.js version in examples/lib to 0.4.24"

islemaster:
Reverts molleindustria/p5.play#86

Opening this for discussion. @cpirich found https://github.com/processing/p5.js/issues/1390 today.  There's reason to believe that most p5.play sketches are unaffected because the camera (which gets enabled at the beginning of every draw loop) does a push/pop that masks the effect of the bug.  However, sketches that call `camera.off()` and make subsequent changes to the render state will find those changes (including transform changes) preserved into the beginning of the next draw call.

I suggest we might want to downgrade our p5.js to a version without this bug, and upgrade again when a fix comes out.


islemaster:
Note: p5.js has released v0.5.0 which supposedly includes a fix for this problem, but the fix doesn't work.  We'll need to wait for v0.5.1?

toolness:
Oh, weird! Sure, I'm fine with reverting the change and then upgrading once a working fix comes out. Thanks!

PR #90, opened 2016-05-17
fix drawFrame bug

henrylake:


islemaster:
Hi there!  Can you provide a little context on the bug you're fixing?

henrylake:
Hi! 

Just a very small thing.

When I tried to draw a single frame of a sprite sheet using the drawFrame method but specified the frame to draw using a numeric index rather than a string name, the whole sprite sheet would get drawn. 

Hopefully this should fix that issue and let you use a number to draw a single frame.

islemaster:
Restarted Travis build now that the linting false positive has been fixed.

islemaster:
Beautiful!  Tried this out myself and the fix looks good.  Thanks for catching this!  I'll merge and then write a test to cover the fix.

henrylake:
No problem, glad to have helped!

Issue #89, opened 2016-05-05
Respect _angleMode before converting degrees to radians


cpirich:
If p5.js `angleMode(DEGREES)` is called, p5play doesn't work properly anymore.

See https://github.com/code-dot-org/code-dot-org/pull/8243

cpirich:
PR now available with fix: https://github.com/molleindustria/p5.play/pull/91

PR #91, opened 2016-05-18
respect _angleMode from p5.js

cpirich:
Resolves https://github.com/molleindustria/p5.play/issues/89

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -1171,7 +1171,12 @@ function Sprite(pInst, _x, _y, _w, _h) {
         if(this.collider instanceof AABB)
         {
         //scale / rotate collider
-        var t = radians(this.rotation);
+        var t;
+        if (pInst._angleMode === pInst.RADIANS) {
+          t = radians(this.rotation);
+        } else {
+          t = this.rotation;
+        }
 
         if(this.colliderType === 'custom')
           {
@@ -1469,7 +1474,11 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
       translate(this.position.x, this.position.y);
       scale(this.scale*dirX, this.scale*dirY);
-      rotate(radians(this.rotation));
+      if (pInst._angleMode === pInst.RADIANS) {
+        rotate(radians(this.rotation));
+      } else {
+        rotate(this.rotation);
+      }
       this.draw();
       //draw debug info
       pop();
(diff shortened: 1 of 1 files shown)


cpirich:
@islemaster

lib/p5.play.js
```diff
@@ -1171,7 +1171,12 @@ function Sprite(pInst, _x, _y, _w, _h) {
         if(this.collider instanceof AABB)
         {
         //scale / rotate collider
-        var t = radians(this.rotation);
+        var t;
+        if (pInst._angleMode === pInst.RADIANS) {
+          t = radians(this.rotation);
+        } else {
+          t = this.rotation;
+        }
```
islemaster: This if/else happens several times. Could we extract a helper for it?  It might have to have a clumsy name like `degreesToAngleModeAngle`?

islemaster:
I think this _does_ resolve #89 but I'm not sure it respects p5's angleMode as much as compensates for it.  Am I correct in understanding that, given these changes, no matter what the angleMode is set to, the p5.play methods expect angles in degrees?

lib/p5.play.js
cpirich: I thought about a helper, but it will be a little tricky because it would depend upon the `pInst` variable, with is actually different for each object, so we'd need to create a helper for both the `Sprite` class and the `Animation` class.

Ultimately, I think we should remove this code once p5.play supports either angle mode (see comment below)

cpirich:
Yes, this is the short-term fix. p5.play still only supports `DEGREES` mode, but at least it can work while p5.js is in `DEGREES` mode.

The next step is to make the p5.play APIs support either `angleMode`, but that is a breaking change because the default value will become `RADIANS`.

islemaster:
Note to self: :cloud: :zap: I should fix the linting error and re-run the CI for the open pull requests.

islemaster:
Okay, I think I'm good with the short-term fix; that is, to keep p5.play working even when angleMode is changed, without making a breaking change to the p5.play API - and we can consider the API breaking change later. :+1: 

I'm going to try and fix the CI build and then get this merged.

islemaster:
Ha!  The linting error is caused by a bug in eslint, introduced in 2.10.0.  I'll update our dependency.

islemaster:
Restarted Travis build now that the linting false positive has been fixed.

PR #95, opened 2016-05-19
Add SpriteSheet tests that cover drawFrame bug.

islemaster:
Adding tests that show a couple of different ways to initialize a SpriteSheet, and some tests showing that drawFrame successfully maps to a p5.image call given a few different ways to call it.  One of those ways is by frame index, which was the bug fixed by #90.  I undid that fix and confirmed that the new tests would have caught the bug.

@henrylake please take a look!

```
  SpriteSheet
    ✓ can be instantiated via constructor 
    ✓ can be instantiated via loadSpriteSheet() 
    ✓ can be initialized with two arguments (image, frames) 
    ✓ can be initialized with four arguments (image, frame_width, frame_height, num_frames) 
    drawFrame
      ✓ draws the expected frame to the destination coordinates 
      ✓ can refer to frame by index 
      ✓ can refer to frame by string name 
      ✓ default destination width and height are source width and height 


  8 passing (19ms)
```

### DIFF
--- a/test/unit/spritesheet.js
+++ b/test/unit/spritesheet.js
@@ -20,4 +20,109 @@ describe('SpriteSheet', function() {
 
     expect(sheet).to.be.an.instanceOf(pInst.SpriteSheet);
   });
+
+  it('can be initialized with two arguments (image, frames)', function() {
+    var image = new p5.Image(200, 160, pInst);
+    var frames = [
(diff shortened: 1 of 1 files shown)


henrylake:
Looks great thanks for the quick response! 

Method seems to be working perfectly now.

PR #96, opened 2016-05-23
lock rotateToDirection bidirectionally, make angle optional on setSpeed (reopened)

cpirich:
(Previously opened as https://github.com/molleindustria/p5.play/pull/85 - now with additional merges applied)

code.org is using p5.play.js as part of a new computer science curriculum. We would like to change the behavior of the Sprite object's rotateToDirection property to make it lock the rotation property with the velocity angle bi-directionally (the current behavior is to ignore the rotation property in this mode and continually overwrite it with the angle of the velocity vector). This change is accomplished by implementing a custom "setter" for the rotation property that updates the velocity vector when rotateToDirection is true. The internal (private) storage for the rotation property value is now stored in the rotation_ property.

In addition, we would like to change the second parameter of the Sprite object's setSpeed function (the angle) to be optional, so that the speed alone can be changed. (Omitting the angle parameter is essentially the same as calling sprite.setSpeed(<new value>, sprite.getDirection()), except we have elected to default the angle to the rotation property when the velocity vector is 0,0). This allows the student to enable rotateToDirection, then freely call setSpeed() with a speed parameter only and set rotation to change direction. The sprite can be stopped in this mode by calling setSpeed(0) and restarted along the same angle by calling setSpeed() with a positive speed value.

The rotationSpeed property will now continue to be respected even while in rotateToDirection mode. To preserve the previous behavior of always locking the rotation property to the velocity angle, the rotation property should not be modified and the rotationSpeed property should be set to zero.


islemaster:
This looks good!

PR #97, opened 2016-07-12
Create internal copy of width and height in Sprite class

caleybrock:
cc: @islemaster 

Create an internal copy of width and height in Sprite class to allow for modifications to with and height property. This should have no affects on p5.play functionality, but for Code Studio Game Lab, we want to change the way students interact with width, height and scale. This PR, along with this Game Lab PR (https://github.com/code-dot-org/code-dot-org/pull/9387) create that functionality. 

In Game Lab, `width` and `height` properties should work like this:
* Sprite starts out as 100x100
* User sets animation for sprite, which is 150, 100.  [sprite.scale: 1; sprite.width: 150; sprite.height: 100]
* User sets scale of sprite to 2.  This updates the width/height. [sprite.scale: 2; sprite.width: 300; sprite.height: 200]
* User sets the sprite width to 260.  [sprite.scale: 2; sprite.width: 260; sprite.height: 200]
* User sets scale of sprite back to 1.  [sprite.scale: 1; sprite.width: 130; sprite.height: 100]
* User sets animation for sprite to something else (say a 50x50 image) and sets the scale again to 2.  [sprite.scale: 2; sprite.width:100; sprite.height:100]
* User sets animation back to the first one. [sprite.scale: 2; sprite.width:300; sprite.height:200]

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -1015,11 +1015,49 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * Width of the sprite's current image.
   * If no images or animations are set it's the width of the
   * placeholder rectangle.
+  * Used internally to make calculations.
+  *
+  * @property internalWidth
+  * @type {Number}
+  * @default 100
+  */
+
+  this.internalWidth = _w;
+
+  /**
+  * Height of the sprite's current image.
+  * If no images or animations are set it's the height of the
+  * placeholder rectangle.
+  * Used internally to make calculations.
+  *
+  * @property internalHeight
+  * @type {Number}
+  * @default 100
+  */
+
+  this.internalHeight = _h;
+
+  /**
+  * Width of the sprite's current image.
+  * If no images or animations are set it's the width of the
+  * placeholder rectangle.
   *
   * @property width
   * @type {Number}
   * @default 100
   */
+
+  Object.defineProperty(this, 'width', {
+    enumerable: true,
+    configurable: true,
+    get: function() {
+      return this.internalWidth;
+    },
+    set: function(value) {
+      this.internalWidth = value;
+    }
+  });
+
   if(_w === undefined)
     this.width = 100;
   else
@@ -1034,6 +1072,18 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * @type {Number}
   * @default 100
   */
+
+  Object.defineProperty(this, 'height', {
+    enumerable: true,
+    configurable: true,
+    get: function() {
+      return this.internalHeight;
+    },
+    set: function(value) {
+      this.internalHeight = value;
+    }
+  });
+
   if(_h === undefined)
     this.height = 100;
   else
@@ -1048,7 +1098,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * @type {Number}
   * @default 100
   */
-  this.originalWidth = this.width;
+  this.originalWidth = this.internalWidth;
 
   /**
   * Unscaled height of the sprite
@@ -1059,7 +1109,7 @@ function Sprite(pInst, _x, _y, _w, _h) {
   * @type {Number}
   * @default 100
   */
-  this.originalHeight = this.height;
+  this.originalHeight = this.internalHeight;
 
   /**
   * True if the sprite has been removed.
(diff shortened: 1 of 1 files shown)


islemaster:
Hey @toolness, this is @caleybrock who's just joined me in working on our Game Lab tool.  We are working on making `Sprite.width` and `Sprite.height` work differently on our end, and this refactor helps us do that.

I think it's also a positive step for p5.play in isolation, as there are some funny inconsistencies around the meaning of `width` and `height` already.  For example, these properties represent unscaled dimensions of a sprite that has no animation assigned (i.e. the box drawn in the sprite's place ends up with width `sprite.width * sprite.scale`) but a sprite with an animation and a default collider expects `width` to represent the _prescaled_ width of the sprite (which gets 'enforced' in `Sprite.prototype.update()`).  This PR shouldn't change any of those behaviors, but breaking the stored state apart from the interface layer is the first step toward making careful changes to those assumptions later.

Agree? Disagree?  How do you feel about this change?

lib/p5.play.js
```diff
@@ -1015,11 +1015,49 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
   * If no images or animations are set it's the width of the
   * placeholder rectangle.
+  * Used internally to make calculations.
+  *
+  * @property internalWidth
+  * @type {Number}
+  * @default 100
+  */
+
+  this.internalWidth = _w;
```
islemaster: We've started using a leading underscore and the `@private` annotation to denote private/internal properties; we should probably do that here.  Example: https://github.com/molleindustria/p5.play/pull/97/files#diff-84d79f68ce33ce29426e52fd37061b4eR921

lib/p5.play.js
```diff
@@ -1015,11 +1015,49 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
+
+  /**
+  * Height of the sprite's current image.
+  * If no images or animations are set it's the height of the
+  * placeholder rectangle.
+  * Used internally to make calculations.
```
islemaster: We've left the descriptions of these internal properties very similar to the public properties; it might be a useful exercise to work out what's _really_ represented by this value in different situations (e.g. the "scaled with an animation and collider, unscaled otherwise" stuff) and document it here.

lib/p5.play.js
```diff
@@ -1015,11 +1015,49 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
+  /**
+  * Width of the sprite's current image.
+  * If no images or animations are set it's the width of the
+  * placeholder rectangle.
   *
   * @property width
   * @type {Number}
   * @default 100
   */
+
```
islemaster: Whitespace between doc comment and the new property.

islemaster: We should locally regenerate the documentation (which is built from these YUIDoc comments) and make sure the width and height properties still show up normally when they're defined this way.  I think they will be because YUIDoc only looks at comments, not code... for that matter, we may want to un-YUIDoc the comments for the internal properties, since we don't want the documented on the public API (and there's a bug with the YUIDoc and private properties, I think).

islemaster:
@caleybrock please review recommended whitespace and documentation changes.  Otherwise this looks good to me.

islemaster:
I did manual validation of all the sprite examples and everything looks unchanged.

caleybrock:
Cool, updated some documentation and made a name change. Thanks!

lib/p5.play.js
caleybrock: I don't totally understand the changes to with scale to these properties, and right now they act the same, so I'm not sure what to differentiate between. The biggest difference is the user of the library should interact with `width` and internally we use `_internalWidth`

toolness:
Hey, sorry it took me so long to get to this!  I think this looks like a fine change... I'm curious if it makes sense to document the possibility of altering the behavior of `Sprite.width` and `Sprite.height` in the way you've done in Game Lab, so as to provide some context around the change?

Anyways, just a thought, since taken out of context it doesn't seem obvious why this refactoring is needed, but perhaps I'm just being dense.  :grinning:   Feel free to merge!

islemaster:
That's a good idea.  @caleybrock maybe you could write something up above the yuidoc comment for the `width` property about how the width/height properties are designed to be extensible if a different behavior (like always representing scaled width) is desired.  I feel like we don't want this generated as part of the API docs (which are mostly targeted at a student user of vanilla p5.play) but it should be right there in the source for anyone who goes looking for those properties.

caleybrock:
@islemaster sorry for the late reply - didn't see this. I added an example of how to modify width, but I'm not sure why this would have failed tests. The failure seems to be unrelated.

lib/p5.play.js
```diff
@@ -1037,6 +1037,22 @@ function Sprite(pInst, _x, _y, _w, _h) {
   */
   this._internalHeight = _h;
 
+  /*
+   * _internalWidth and _internalHeight are used for all p5.play
+   * calculations, but width and height can be exteneded. For example,
```
islemaster: Typo: s/exteneded/extended

islemaster:
Weird... I'll see if I can track it down.

PR #98, opened 2016-07-13
Update collision calculations to use current width and height

caleybrock:
Updates collisions with a default collider to use the current width and height. 

@islemaster @toolness  This fixes a case where colliders weren't updating properly when a sprite's width/height got changed by a user. Also `originalHeight` and `originalWidth` are no longer used internally. Should we go ahead and get rid of it or just keep it incase others are using it? Thoughts? 

For the "Collisions - overlap, collide, displace" example, I removed the animations and set the rectangles to have a new width and height: `asterisk.width = 200;` and `cloud.height = 120;`

Before fix: In debug mode you can see that the collision rectangle remains at 100x100 which is the original width set for the sprite, instead of updating to 200x100.
screen shot 2016-07-12 at 5 24 06 pm

After fix: Collision rectangle matches the sprite size.
screen shot 2016-07-12 at 5 22 03 pm

Note - I think once PR (https://github.com/molleindustria/p5.play/pull/97) gets updated, we should be using the `_internalWidth` and `_internalHeight` to update collision rectangles.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -1276,10 +1276,10 @@ function Sprite(pInst, _x, _y, _w, _h) {
           }
         else if(this.colliderType === 'default')
           {
-          this.collider.extents.x = this.width * abs(this.scale) * abs(cos(t)) +
-          this.height * abs(this.scale) * abs(sin(t));
-          this.collider.extents.y = this.width * abs(this.scale) * abs(sin(t)) +
-          this.height * abs(this.scale) * abs(cos(t));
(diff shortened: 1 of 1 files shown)


islemaster:
`originalWidth` and `originalHeight` are documented as public properties, so removing them would be a breaking API change (even though we don't have a versioning/release process yet, we try to avoid those).  We should probably leave them for now.

caleybrock:
Updated with master and used internal values of width and height since that was already merged.

PR #99, opened 2016-07-14
Fix width and height accesses from frames for animations

caleybrock:
@islemaster @toolness 
Animations created with spritesheets were returning NaN for `width` and `height`. This fixes it so that users can access a sprite's height and width.

What's next: write some unit tests for animations. I'll submit another PR with these.


islemaster:
Concur.  This definitely fixes a bug, I'm gonna go ahead and merge.  Looks like Travis is having issues today, but I verified linting and tests locally.

PR #100, opened 2016-07-14
Make getDirection return degrees even if p5 angleMode is DEGREES

islemaster:
Before if p5 angleMode was DEGREES the getDirection result got double-converted, which led to nonsense values.  Unit-tested.


islemaster:
@caleybrock please CR as well.

islemaster:
Travis build failed on npm install because phantomjs CDN was unreachable.  Retrying...

caleybrock:
👍

PR #102, opened 2016-07-16
Upgrade ESLint dependency

### LINKED ISSUE #93: Support ESLint >= 2.10.3 (if regression is fixed)
We dropped support for ESLint >= 2.10.0 in #92 due to a [regression causing a false positive when linting our code](https://github.com/eslint/eslint/issues/6173).  The [fix for the regression has been merged to master](https://github.com/eslint/eslint/pull/6192) so it should probably go out with 2.10.3.   Once that happens, we should update our dependency to allow the newer versions with the fix.

islemaster:
Fixes #93.

Update from `>=2.5.3 <2.10.0` to `^3.1.0` which is the latest version.  The major breaking change with this update is [dropping support](https://github.com/eslint/eslint/blob/master/docs/user-guide/migrating-to-3.0.0.md#dropping-support-for-nodejs--4) for Node.js < 4 which shouldn't be an issue for this repo (as ESLint points out, Node 4 is the current LTS version).  I've added the engines field in package.json to document the dependency on Node >= 4.

Fixes the one new linting violation caught after the upgrade (and I'm actually surprised this one wasn't caught _before_ it).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,2 @@
 language: node_js
+node_js: 4
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -128,7 +128,9 @@ var round = p5.prototype.round;
 * @type {Group}
 */
 
-defineLazyP5Property('allSprites', function() { return new Group(); });
+defineLazyP5Property('allSprites', function() {
+  return new Group();
+});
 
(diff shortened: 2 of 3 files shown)


toolness:
Nice, this looks good to me. Thanks @islemaster !

PR #103, opened 2016-07-23
Fix width and height overriding

caleybrock:
@islemaster @toolness Hi! I was running into some issues with using p5.play to modify and get a sprite's width and height when it has an animation. The first item is a bug fix, but I also wanted to increase flexibility of p5.play so that logic could be more easily modified by users of the library so I added a few new functions as well.

- Initialize `width` and `height` when adding an animation
- Factor out updating sizes for animations outside of the update function
- Create `_getScaleX` and `_getScaleY` (currently as getters for scale) which opens the opportunity to have a horizontal scale and a vertical scale that are different.
- Added a test for an animation with frames that have different sizes. The width and height should update when each frame is updated. This happens in the helper `this._syncAnimationSizes` now

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -1129,23 +1129,21 @@ function Sprite(pInst, _x, _y, _w, _h) {
     //the animation wasn't loaded. if the animation is not a 1x1 image
     //it means it just finished loading
     if(this.colliderType === 'default' &&
-      animations[currentAnimation].getWidth() !== 1 &&
-       animations[currentAnimation].getHeight() !== 1
-      )
+      animations[currentAnimation].getWidth() !== 1 && animations[currentAnimation].getHeight() !== 1)
     {
-    this.collider = this.getBoundingBox();
-    this.colliderType = 'image';
-    this.width = animations[currentAnimation].getWidth()*abs(this._getScaleX());
-    this.height = animations[currentAnimation].getHeight()*abs(this._getScaleY());
-    //quadTree.insert(this);
+      this.collider = this.getBoundingBox();
+      this.colliderType = 'image';
+      this.width = animations[currentAnimation].getWidth()*abs(this._getScaleX());
+      this.height = animations[currentAnimation].getHeight()*abs(this._getScaleY());
+      //quadTree.insert(this);
     }
 
     //update size and collider
     if(animations[currentAnimation].frameChanged || this.width === undefined || this.height === undefined)
     {
-    //this.collider = this.getBoundingBox();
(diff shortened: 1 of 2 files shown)


lib/p5.play.js
```diff
@@ -1120,6 +1120,35 @@ function Sprite(pInst, _x, _y, _w, _h) {
...
+    //quadTree.insert(this);
+    }
+
+    //update size and collider
+    if(animations[currentAnimation].frameChanged || this.width === undefined || this.height === undefined)
+    {
+    //this.collider = this.getBoundingBox();
+    this.width = animations[currentAnimation].getWidth()*abs(this._getScaleX());
+    this.height = animations[currentAnimation].getHeight()*abs(this._getScaleY());
+    }
```
islemaster: Indentation looks a little funky through here, could clean up formatting?

test/unit/sprite.js
```diff
@@ -76,5 +76,55 @@ describe('Sprite', function() {
       checkDirectionForVelocity([1, 1], 45);
       checkDirectionForVelocity([0, 1], 90);
     });
+
+  });
+
+  describe('demension updating when animation changes', function() {
```
islemaster: Spelling: s/demension/dimension

islemaster:
Couple of nits but otherwise this looks like a great refactor to me and has tests! :tada:  @toolness thoughts?

caleybrock:
Updated code review and fixed conflicts with master.

islemaster:
Legitimate failure on travis?
```
 1) Example sketches sprites_with_sheet.js runs for 3 frames:
    TypeError: 'undefined' is not an object (evaluating 'this.images[frame].width') @ file:///home/travis/build/molleindustria/p5.play/lib/p5.play.js:3802
```

caleybrock:
looking 👀

lib/p5.play.js
```diff
@@ -3762,22 +3794,32 @@ function Animation(pInst) {
...
   *
   * @method getWidth
   * @return {Number} Frame width
   */
   this.getWidth = function() {
-    return this.images[frame].width;
+    if (this.images[frame]) {
+      return this.images[frame].width;
+    } else {
+      return 1;
```
islemaster: Comment about when/why we would return 1 here?

islemaster: Never mind, comment is in method documentation.

caleybrock:
@islemaster @toolness any additional thoughts on this? good to merge?

islemaster:
This looks great.  Thanks Caley! :tada:

PR #106, opened 2016-07-29
Drag fix in instance mode

islemaster:
Follow-up to https://github.com/molleindustria/p5.play/commit/8151d9b1d7e4923b07956fd9646036536655d23a that makes that change work for instance mode.

Also adds the related dragging behavior to the mouse events example, and adds tests for Sprite mouse callbacks:
```
    mouse events
      ✓ mouseIsOver property represents whether mouse is over collider 
      onMouseOver callback
        ✓ calls onMouseOver when the mouse enters the sprite collider 
        ✓ does not call onMouseOver when the mouse moves within the sprite collider 
        ✓ calls onMouseOver again when the mouse leaves and returns 
      onMouseOut callback
        ✓ calls onMouseOut when the mouse leaves the sprite collider 
        ✓ does not call onMouseOut when the mouse moves outside the sprite collider 
        ✓ calls onMouseOut again when the mouse returns and leaves 
      onMousePressed callback
        ✓ does not call onMousePressed if the mouse was not over the sprite 
        ✓ calls onMousePressed if the mouse was pressed over the sprite 
        ✓ calls onMousePressed if the mouse was pressed outside the sprite then dragged over it 
      onMouseReleased callback
        ✓ does not call onMouseReleased if the mouse was never pressed over the sprite 
        ✓ calls onMouseReleased if the mouse was pressed and released over the sprite 
        ✓ calls onMouseReleased if the mouse was pressed, moved over the sprite, and then released 
        ✓ does not call onMouseReleased on mouse-out if mouse is still down 
        ✓ does not call onMouseReleased on release if mouse has left sprite 
```

### DIFF
--- a/examples/mouseEvents.js
+++ b/examples/mouseEvents.js
@@ -3,6 +3,7 @@
 
 var asterisk;
 var ghost;
+var draggedSprite;
 
 function setup() {
   createCanvas(800, 400);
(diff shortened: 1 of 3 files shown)


caleybrock:
👍 thanks brad!

islemaster:
I'm going ahead and merging this since it fixes an instance-mode bug.  @molleindustria mentioned there's probably additional work that should be done to suppport dragging sprites, but I suspect that should be done in a separate change (and extend the mouse callback tests introduced here).

PR #109, opened 2016-08-29
_getScaleX and _getScaleY missing from default collisions

caleybrock:
Found a missing case in default collisions where we are referencing `scale` instead of `getScaleX()` and `getScaleY()`. (See the 10 lines above where scale functions are called properly).


caleybrock:
@islemaster @toolness This bug was my bad - noticed when updating the p5.play library in code-dot-org. I had fixed the bug in our repository, but forgot to update it here.

islemaster:
Nice catch!  Thanks!

PR #114, opened 2016-10-17
Make collide() and displace() affect velocity as inelastic collisions

islemaster:
This is a proposal to change the behavior of `Sprite.collide()` and `Sprite.displace()`.  We'll be modifying this behavior in our own application and thought it might be more generally popular, as we think it leads to less 'surprising' behavior when one begins using p5.play.

The key change is that `collide()` and `bounce()`, which previously had no effect on a sprite's `velocity` property, would now cause the displaced sprite to adopted the other's velocity along the axis of displacement - simulating an inelastic collision with an immovable object.  Before this change, these methods have no effect on the sprite's velocity which can lead to a surprising mismatch between a sprite's visual behavior and its internal state.

# Examples

Here are some examples of how this changes sprite behavior.  Note, these examples are from our Gamelab UI, but the behaviors shown apply to vanilla p5.play too.

## Example 1: Squares colliding
Before when calling `A.collide(B)` square B would affect square A's position, but its velocity would be untouched, so the squares would resume on their original paths once they were no longer colliding:

collide

After this change, the displaced block adopts the displacing block's velocity so it continues on its new path after the separate:

collide

And here's a before/after of `displace()` which is the same with the roles of the two sprites flipped:

displace
displace

## Example 2: Circles colliding

In the before case this is much like the two squares colliding - the circles move apart then continue in their original directions.

collide

Now it's markedly different - not only does the circle change direction, it's deflected at nearly 90 degrees because the collision normal is no longer axis-aligned.

collide

Again, here's the before/after for `displace()`:

displace
displace

# Example 3: Platforming

This is the case that got the most attention from our team.  If you set up simple gravity and try to use `collide()` to keep a sprite on a platform, that sprite will continue to accelerate while it sits on the platform causing it to drop like a rock when it goes over the edge:

collide

But now its velocity is correctly arrested, giving it a graceful arc as it leaves the platform.  It's also worth noting that the ball's `velocity.x` increases a little now as it slides off the edge and the collision normal changes.

collide

I'd share the displace case here but it's not particularly interesting :grin::

displace

# Testing

I broke up the existing collision unit tests into separate files since they were getting particularly long, and added several new tests for the new behaviors.  Note, if there's interest in splitting up the tests this way but _not_ in the new behavior I'd be happy to open a PR just splitting up the tests and pinning the existing behavior.

I also manually ran through the set of examples that come with p5.play.  It looks like they all still work as expected, though I don't think any of them exercise this particular behavior well.


islemaster:
Looking at failed tests...

molleindustria:
I agree that Example 3 is problematic. The rationale was to have bounce() to resolve elastic collisions and collide/displace to resolve overlaps in a non-physics-based situation, but I don't see a problem in changing velocities if they are present.

I see a problem in transferring forces to objects that are not already moving though velocity. It would be easy to assume that all the sprites have a rigid body and this is a simplified version of box2D but the reality is that the 90% of 2D games in which things hit each other (eg atari era classics, platformers, top down rpgs, roguelikes, shmups...) simply don't use "real" forces and elastic collisions.

Maybe the .mass property set at -1 (default) can be used to determine if the body is affected by forces, maybe there can be a .rigidbody boolean that automatically activates if a force is applied but is otherwise false until changed.

islemaster:
Having a negative mass or a boolean to check whether a sprite is participating in the simulation might create more problems than it solves.  In some ways this is what the `immovable` property is for, but I'd hate to overload that property.

For future reference, can you help me out with the original intent of the various collision functions?  Here's how they are commented in the library:

* `collide()`: If the overlap is positive the current sprite will be displace by the colliding one in the closest non-overlapping position.
* `displace()`: If the overlap is positive the current sprite will displace the colliding one to the closest non-overlapping position
* `bounce()`: If the overlap is positive the sprites will bounce affecting each other's trajectories depending on their .velocity, .mass and .restitution

Here's my current understanding:

* `collide()` and `displace()` will only move a sprite out of collision, changing its position (and its `touching` property) but never its velocity, so there's effectively no 'physics' going on.
  * They always treat one of the actors as 'immovable'.
  * It seems like these two functions are the same thing with actors reversed.  Is that true?  They're currently implemented as separate code paths, I'll investigate doing some cleanup here if they're supposed to be the same.
* `bounce()` actually provides simplified physics, taking velocity/mass/restitution/immovable into account and changing sprite velocities in addition to positions.

It sounds like the behavior I'm trying to implement on our end is actually `bounce` with restitution set to zero and one sprite temporarily immovable.  Maybe I'll re-implement in those terms instead of trying to modify it here.  Any interest in still having the tests broken up like this?

islemaster:
Here's a before/after example that I think illustrates your point about not adding velocity to an object at rest.  You're saying that the 'before' behavior is the more desirable one here?

**Before:**
transfer_before

**After:**
transfer_after

molleindustria:
Yes, collide and displace are basically the same thing and I think I remembered them being kinda messy. The idea was to separate the collision detection from the overlap resolution and the physics response (bounce() which may not be a good name) because they are often used separately. I'm definitely in favor of a simplification or reimplementation, I'd only avoid assuming that all the sprites participate in the physics system and that the sprites are always moved by forces.

islemaster:
I actually think this behavior - which is 'after' with `s2.friction = 0` (100% friction) - is better than the 'before' case, because `s2`'s velocity is modeled correctly while it's being pushed.

transfer_after_with_friction

But it's a non-issue if the idea is just that `collide()`/`displace()` are not really supposed to affect sprite momentum.  I'm having trouble recreating this behavior with `bounce()` though - I can investigate that separately.

islemaster:
Okay.  So `a.bounce(b)` assumes that both `a` and `b` are participating in the physics system, but none of the other methods do/should.  As an idealized hierarchy:

* `a.overlap(b)` checks for intersection of colliders.
* `a.displace(b)` checks for intersection _and_ displaces `b` out of `a`.
* `a.bounce(b)` checks for intersection, displaces, _and_ modifies velocities of both `a` and `b` using a simple physics model.

I like that; I'll work on cleanup that follows this model and get back to you.

I think our team here suspects students will not use `displace` much, and in reality what we're asking for are variations on `bounce` with certain assumptions.  For instance, we've already introduced `a.bounceOff(b)` in our version, which is just `bounce` where `b` is implicitly immovable.  I'll communicate that back here, it may help clear up what we're doing.

Thank you for all your help on this!

molleindustria:
Yeah, the last behavior is definitely more desirable. 
Then I would ditch displace and bounce and simply have a collide() that resolves both restitution and overlaps. If you want it to push a thing you set maximum friction, if you it want to be on platforms you set them as immovable, if you want it to be moved you set widely different masses. If you don't control movements via velocity (eg. position updated by mouse position, or non continuous discrete leaps) then collide would just separate the overlapping objects.

Also did I really set max friction as 0? Ewww, should it be flipped 0=0% 1=100%?

islemaster:
I'm closing this PR since it sounds like we've decided against the changes it proposes.  I'll follow up with some smaller changes:

* Collision unit tests (over _existing_ behavior) in separate files.
* Clean up existing implementation, in particular unifying code paths for `collide` and `displace`.

I'd be in favor of flipping the `friction` property, if you're okay with breaking changes.  @toolness and I have been pretty cautious about that, since we don't have a system for publishing a new version with release notes, but I'd be surprised if many people were auto-updating p5.play through some package manager - I imagine most users build on one version and rarely if ever upgrade mid-project.

Issue #104, opened 2016-07-26
createGraphics does not work with p5.play in instance mode. 


arorag2396:
``` js
p.setup = function () {
            canvas = p.createCanvas(1000, 1000);
            pg = p.createGraphics(1000, 1000);
}
```

```
Uncaught TypeError: pInst.createVector is not a function
Camera @ p5.play.js:2387
(anonymous function) @ p5.play.js:2359
get @ p5.play.js:58
p5.Graphics @ p5.js:11555
p5.createGraphics @ p5.js:13288
p.setup @ RTChess.js:58
(anonymous function) @ p5.js:8900
_runIfPreloadsAreDone @ p5.js:8853
_decrementPreload @ p5.js:8862i
mg.onload @ p5.js:17669
```

It works in non instance mode, but seems not be working for instance mode. Am I doing something wrong?

ycatch:
Hi, there.

I tried p5.js and p5.play in instance mode, like this.

	<html>
	<head>
	  <meta charset="UTF-8">
	  <style> body {padding: 50; margin: 0;} </style>
	</head>

	<body>
		<script src="p5.js"></script>
		<script src="p5.play.js"></script>
		<script>
		var s = function(p) {

			var sq;
			var layer;
			p.setup = function() {
				p.createCanvas(800,400);
				p.background(200);
				layer = p.createGraphics(400, 200, 0, 0);
				sq = p.createSprite(400, 200, 50, 50);
			}

			p.draw = function() {
				p.drawSprites();
				p.image(layer);
			}
		}

		var myp5 = new p5(s);
		</script>
	</body>
	</html>

However, I had error message.

    TypeError: pInst.createVector is not a function

When I commented createGraphics out, it worked.

islemaster:
Hi there! `createGraphics` is a method defined by the p5.js library, not specifically by p5.play.  Does this problem only occur when you are using p5.play?  If not, you may want to re-open this issue over at https://github.com/processing/p5.js.

I'll be happy to investigate later today if the problem is specific to p5.play.

arorag2396:
@islemaster Yes, this problem is specific to p5.play. I have used createGraphics couple of times in p5.js, where I didn't use p5.play.
Thanks!

ycatch:
islemaster thank you for your comment.
I posted this issue to p5.js/

https://github.com/processing/p5.js/issues/1635

islemaster:
[...]

Issue #46, opened 2016-03-14
Use 'init' registerMethod() instead of lazily-setting properties


toolness:
Some complexity in supporting instance mode was introduced because there wasn't a way for extension libraries to do initialization when a sketch is first created. This resulted in us having to create a function called `defineLazyReadOnlyP5Property()`, among other things. Some discussion about the complexity is at https://github.com/molleindustria/p5.play/pull/22#discussion_r53211269.

Anyways, since then, I've contributed a new `'init'` `registerMethod()` type to core p5 in https://github.com/processing/p5.js/pull/1291, so going forward, we could potentially use that instead of `defineLazyReadOnlyP5Property()`, which could simplify p5.play's internals a bit.

quinton-ashley:
Interesting, I will look into this.

quinton-ashley:
@toolness I finished what you started and completed this in v2.1.3 :)

quinton-ashley:
@toolness Actually I noticed that doing it that way only kind of worked but `instanceof` did not work because each class was essentially being turned into a separate function which made all the work I did turning them into classes (for better performance) pointless. I improved upon it by putting all the classes inside the `p5PlayInit` function and then storing `this` (the p5 instance) in a variable `pInst` which I could then use inside the classes. It took me a while to figure out but now `instanceof` works.

quinton-ashley:
@toolness v2.2.0 is ready!

CyberSix6:
new at this, please refer to open ones.

quinton-ashley:
@CyberSix6 what?

PR #119, opened 2016-10-20
Fix default camera in instance mode

islemaster:
Attempts to access the lazy-created default `camera` property in instance mode were leading to errors, and it was also interfering with p5's `createGraphics()` [as reported in #104](https://github.com/molleindustria/p5.play/issues/104).  This PR registers an initialization method with p5 [as proposed in #46](https://github.com/molleindustria/p5.play/issues/46) and creates the default camera there instead.  A number of changes were necessary to make this viable, but I think it's worth it, as we can gradually migrate away from lazy-created properties to using the init method for everything.
# p5.js upgrade

I had to upgrade the p5.js version used by the examples; the ability to register an 'init' method was released in [0.4.24](https://github.com/processing/p5.js/releases/tag/0.4.24) and we were on [0.4.21](https://github.com/processing/p5.js/releases/tag/0.4.21).  I went ahead and upgraded all the way to [0.5.4](https://github.com/processing/p5.js/releases/tag/0.5.4).  This means we now depend on p5.js >= 0.4.24, but I'm not sure we've previously recorded that dependency anywhere.  Opinions on where to put it?
# Examples page changes

It seems registering an init method was not working on the examples page because it would load p5.js once on the page, but would eval p5.play with each example - and also because, for some reason, it was clearing all registered methods prior to eval'ing the examples.  I verified that removing this does not cause us to register init too many times.

One nice side effect of this is that you can debug p5.play on the examples page now, which was previously difficult since it was being eval'd.

I think the biggest risk is possible state bleed between examples, but if that's possible I haven't seen it happen yet.  We seem to create a new p5 instance for every example anyway.
# Fix up sprites_with_sheet example

I'm not sure how this was working before - we were using the result of a loadJSON call on another line inside `preload()`, which is not allowed.  To fix, I'm using its result inside a callback - I know it's an advanced technique, but it's probably necessary here if you want to generate a spritesheet in preload() using asynchronously-fetched JSON.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -4,11 +4,12 @@
     <title>p5.play Example</title>
     <script src="lib/jquery.min.js" type="text/javascript"></script>
     <script src="lib/p5.js" type="text/javascript"></script>
+    <script src="../lib/p5.play.js" type="text/javascript"></script>
     <script src="js/ace/ace.js"></script>
     <script src="player.js"></script>
     <link rel="stylesheet" href="js/bootstrap/bootstrap.min.css">
     <script src="js/bootstrap/bootstrap.min.js"></script>
-    
+
     <link rel="stylesheet" type="text/css" href="style.css">
 
     <style>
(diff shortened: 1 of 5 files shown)


molleindustria:
I can update the examples on the "official" page if things changed. 
To this date I still have no idea of how instance mode works.

Also, should I set up a github web page that can be updated collectively (right now it's just static stuff on my server)?

islemaster:
Note to self: Although tests are passing, I'd like to clean up the new warning `You just changed the value of "camera", which was a p5 function. This could cause problems later if you're not careful.` before merging.

islemaster:
@molleindustria we are already serving these at [molleindustria.github.io/p5.play](http://molleindustria.github.io/p5.play/), such that anyone with write access to the repository should be able to update it by running [update-gh-pages.sh](https://github.com/molleindustria/p5.play/blob/master/update-gh-pages.sh). [Back in March @toolness suggested](https://github.com/molleindustria/p5.play/issues/57#issuecomment-202535945) a way to have p5play.molleindustria.org point to the version at github.io, using a CNAME DNS record.

It would be a small step now to get that environment to auto-deploy any time `master` is updated - I do exactly that in [this other repo](https://github.com/islemaster/hitotoki) to keep the ["play in browser" feature](http://islemaster.github.io/hitotoki/play.html) current with the master branch.

islemaster:
Regarding the warning, I'm not sure I should fix in this PR; p5.play is legitimately replacing the [p5.js method `camera`](http://p5js.org/reference/#/p5/camera), [present since 0.4.10](https://github.com/processing/p5.js/commit/ac42e37dcc43340c6d56bef1c8feb35c8b4d591e) (September 2015) with its own default camera ([not in the docs but used in the camera.js example](http://p5play.molleindustria.org/examples/index.html?fileName=camera.js)), present at least since the initial commit in June 2015 :frowning_face:.  There just hasn't been a warning until now.

islemaster:
Haven't had a lot of feedback on this change, but it definitely fixes issue so I'm merging.

PR #120, opened 2016-10-20
Flip friction value

islemaster:
Proposed breaking change: Reverse the meaning and default value of the Sprite `friction` property so that zero means 'no friction' and 1 means 'full friction.'

Suggested by @molleindustria in https://github.com/molleindustria/p5.play/pull/114#issuecomment-254623761, and I agree this is more intuitive.
# New Tests

```
  Sprite
    friction
      ✓ has no effect on update() when set to 0 
      ✓ reduces velocity to zero on update() when set to 1 
      axis-aligned
        ✓ cuts velocity in half each update when set to 0.5 
        ✓ cuts velocity to one-quarter each update when set to 0.75 
      not axis-aligned
        ✓ cuts velocity in half each update when set to 0.5 
        ✓ cuts velocity to one-quarter each update when set to 0.75 
```

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -827,14 +827,15 @@ function Sprite(pInst, _x, _y, _w, _h) {
 
   /**
   * Friction factor, reduces the sprite's velocity.
-  * The friction should be close to 1 (eg. 0.99)
-  * 1: no friction
+  * The friction should be close to 0 (eg. 0.01)
+  * 0: no friction
+  * 1: full friction
   *
   * @property friction
   * @type {Number}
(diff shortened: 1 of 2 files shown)


islemaster:
For one insane moment I wondered if it was correct to apply friction to each axis independently; threw together [this sample](https://gist.github.com/islemaster/da2708d3340b0f0b4c988cd44d5757f9) to convince myself the existing code is right :laughing:.

islemaster:
Going ahead with this since @molleindustria brought it up in the first place. :grin:

PR #122, opened 2016-10-27
In docs, fix filtering class table of contents by member access

islemaster:
I figured it out!

out

Entries in the table of contents on class pages now show/hide with the options checkboxes just like the actual entries do.

Now I won't feel so bad documenting private methods.


islemaster:
Closing, this change is absorbed into #123.

Issue #125, opened 2016-12-09
SpriteSheets Broken in latest P5.play library


gtoast:
animation() call fails to animate.  Even the "Animation - sprite sheet" demo example fails to actually animate the spritesheet. 

Prints message:  Warning undefined frame 0 p5.js:12477

adibbehjat:
+1

molleindustria:
I encountered that too and I'm using an old version in my class. I'll try to investigate but these are not my changes...

gtoast:
I think I found the source of this bug. I'll attempt a patch in a couple hours.

SnipedCode:
Hi, I get very strange result with SpriteSheets. I can't get it work, I copy paste the exemple, image everything and still.... I saw a fix I changed the original file according to the mentioned modification but again, it's not working.. after a day trying and trying... is someone aware of a problem loading SpriteSheets ? please help me understand why I can't load my animation.

galinabee:
I'm having this issue when trying to start up [http://molleindustria.github.io/p5.play/examples/index.html?fileName=animation.js]. Any suggestions?

luyingzhang:
> I'm having this issue when trying to start up [this example](http://molleindustria.github.io/p5.play/examples/index.html?fileName=animation.js) Any suggestions?

restore to the version 0.5.4 can fix:

p5.prototype.image =
  function(img, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight) {
  // set defaults per spec: https://goo.gl/3ykfOq
  if (arguments.length <= 5) {
    dx = sx || 0;
    dy = sy || 0;
    sx = 0;
    sy = 0;
    if (img.elt && img.elt.videoWidth && !img.canvas) { // video no canvas
      var actualW = img.elt.videoWidth;
      var actualH = img.elt.videoHeight;
      dWidth = sWidth || img.elt.width;
      dHeight = sHeight || img.elt.width*actualH/actualW;
      sWidth = actualW;
      sHeight = actualH;
    } else {
      dWidth = sWidth || img.width;
      dHeight = sHeight || img.height;
      sWidth = img.width;
      sHeight = img.height;
    }
  } else if (arguments.length === 9) {
    sx = sx || 0;
    sy = sy || 0;
    sWidth = _sAssign(sWidth, img.width);
    sHeight = _sAssign(sHeight, img.height);

    dx = dx || 0;
    dy = dy || 0;
    dWidth = dWidth || img.width;
[...]

PR #126, opened 2016-12-10
Fixed spritesheet animation

gtoast:
https://github.com/molleindustria/p5.play/issues/125

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -3515,9 +3515,7 @@ function Animation(pInst) {
   else if (frameArguments.length === 1 && (frameArguments[0] instanceof SpriteSheet))
   {
     this.spriteSheet = frameArguments[0];
-    this.images = this.spriteSheet.frames.map( function(f) {
-      return f.frame;
-    });
+    this.images = this.spriteSheet.frames;
   }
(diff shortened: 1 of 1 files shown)


islemaster:
Thanks for putting together a fix! Taking a look now, as there's a good chance I caused this problem :cry:  For easy access, the currently failing tests are:
```
  2 failing

  1) Sprite dimension updating when animation changes animation width and height get inherited from frame:
     expected NaN to equal 30

  2) Sprite dimension updating when animation changes updates the width and height property when frames are different sizes:
     expected NaN to equal 50
```

islemaster:
Looks like the regression was introduced in https://github.com/molleindustria/p5.play/pull/99 and your fix here re-introduces the bug that _that_ PR was trying to fix.  Let's see if we can fix both.

A unit test to cover this regression would be great too.

SnipedCode:
Is this problem fixed? Cause I can't load my SpriteSheets. please help me

islemaster:
Looking at this today.  Sorry it's been a while!

islemaster:
Replaced by #127, which makes this fix but also has changes to avoid re-introducing the old bug.

PR #132, opened 2017-05-19
Fix spritesheets position and offset

cbsrobot:
npm run test throws following errors - but they do not seem related to the change:

group.bounce(group)
Error loading resource file:///Users/me/devel/p5.play/test/fakeImage0001.png

Example sketches
    ◦ animation.js runs for 3 frames: You just changed the value of "camera", which was a p5 function. This could cause problems later if you're not careful.

collisions.js runs for 3 frames: p5 had problems creating the global function "box", possibly because your code is already using that name as a variable. You may want to rename your variable to something else.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -3564,8 +3564,8 @@ function Animation(pInst) {
       {
         if (this.spriteSheet) {
           var frame_info = this.images[frame].frame;
-          pInst.image(this.spriteSheet.image, frame_info.x, frame_info.y, frame_info.width,
-            frame_info.height, this.offX, this.offY, frame_info.width, frame_info.height);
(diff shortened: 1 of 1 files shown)


islemaster:
Thanks for investigating and letting us know! That's a pretty significant breaking change for a patch version. :cry: It sounds like we'll need your change, and possibly others, to provide support for p5.js >= 0.5.5.  I'll take a look at adding support for latest p5.js this weekend.

islemaster:
Hey @cbsrobot! I've integrated your commit here into a PR that makes p5.play fully compatible with the latest version of p5.js.  I'm closing this PR in favor of that one.  Will you review it and let me know what you think?
https://github.com/molleindustria/p5.play/pull/140

PR #140, opened 2017-06-14
Support p5.js >= 0.5.5

### LINKED ISSUE #136: Sprite sheets not working correctly with p5.js >= v0.5.5
I don't know if this is my mistake or if the documentation is behind but I cannot get
the loadSpriteSheet method to work properly. I tried making a json specifying the x and y of the frame, but instead the same frame is chosen and displaced by the x amount. The same thing happens with the short
version where you choose height width and number of frames. the same first frame is chosen and displaced by the x amount.
I couldn't find the issue if it already exists.

islemaster:
A [breaking change](https://github.com/processing/p5.js/issues/1653) ([commit](https://github.com/processing/p5.js/commit/6ac5bc00e101c739e1c423939294a8db51cf3210)) to the argument order of the `image()` function in p5.js 0.5.5 broke spritesheets in p5.play.  This PR makes p5.play compatible with p5.js 0.5.5-0.5.11 (latest as of now).  It _removes_ support for p5.js < 0.5.5; developers wishing to use older p5.js versions will need to use old p5.play versions as well.

Thanks to @cbsrobot and @cesarrodas for raising this issue in https://github.com/molleindustria/p5.play/pull/132 and https://github.com/molleindustria/p5.play/issues/136 respectively.  I've integrated #132 into this PR, replacing it, and this should resolve #136.

I've updated the test matrix to run unit tests against each p5.js version 0.5.5-0.5.11 and against the latest version as well to catch future changes.

I had to blacklist one of the examples from the unit test suite because it uses `loadJSON()`, which was reimplemented with `fetch` in p5.js@0.5.8 and now refuses to load JSON assets from `file://` URLs.  The behavior is not broken. That test still runs and passes when running the test suite in the browser - maybe soon we can update our test runner to run the suite from a local dev server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,11 @@ before_script:
   - 'if [ "$P5_VERSION" ]; then npm install p5@$P5_VERSION && npm run link:p5; fi'
 
 env:
-  - P5_VERSION=0.5.0
-  - P5_VERSION=0.5.1
-  - P5_VERSION=0.5.2
-  #- P5_VERSION=0.5.3 Known issue
-  - P5_VERSION=0.5.4
+  - P5_VERSION=0.5.5
+  - P5_VERSION=0.5.6
+  - P5_VERSION=0.5.7
+  - P5_VERSION=0.5.8
+  - P5_VERSION=0.5.9
+  - P5_VERSION=0.5.10
+  - P5_VERSION=0.5.11
+  - P5_VERSION=latest
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ You can find examples and more information at [p5play.molleindustria.org][].
 
 p5.play provides a Sprite class to manage visual objects in 2D space and features such as animation support, basic collision detection and resolution, sprite grouping, helpers for mouse and keyboard interactions, and a virtual camera. 
 
-p5.play extends [p5.js][], a javascript library (and a community) that aims to make coding accessible for artists, designers, educators, and beginners. If you are not familiar with p5.js, you should start at [p5js.org/tutorials][].
+p5.play extends [p5.js][] (>=0.5.5), a javascript library (and a community) that aims to make coding accessible for artists, designers, educators, and beginners. If you are not familiar with p5.js, you should start at [p5js.org/tutorials][].
 
 ## Development
 
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -3919,8 +3919,8 @@ function SpriteSheet(pInst) {
     }
     var dWidth = width || frameToDraw.width;
     var dHeight = height || frameToDraw.height;
-    pInst.image(this.image, frameToDraw.x, frameToDraw.y,
-      frameToDraw.width, frameToDraw.height, x, y, dWidth, dHeight);
+    pInst.image(this.image, x, y, frameToDraw.width, frameToDraw.height,
+      frameToDraw.x, frameToDraw.y, dWidth, dHeight);
   };
 
   /**
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -3564,8 +3564,8 @@ function Animation(pInst) {
       {
         if (this.spriteSheet) {
           var frame_info = this.images[frame].frame;
-          pInst.image(this.spriteSheet.image, frame_info.x, frame_info.y, frame_info.width,
-            frame_info.height, this.offX, this.offY, frame_info.width, frame_info.height);
+          pInst.image(this.spriteSheet.image, this.offX, this.offY, frame_info.width,
+            frame_info.height, frame_info.x, frame_info.y, frame_info.width, frame_info.height);
         } else {
           pInst.image(this.images[frame], this.offX, this.offY);
         }
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "eslint": "^3.1.0",
     "http-server": "^0.9.0",
     "mocha-phantomjs": "^4.0.1",
-    "p5": "0.5 - 0.5.2 || 0.5.4",
+    "p5": "^0.5.5",
     "yuidocjs": "^0.10.0"
   },
   "engines": {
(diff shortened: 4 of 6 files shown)


GoToLoop:
> ... with fetch in p5.js@0.5.8 and now refuses to load JSON assets from file:// URLs.

Just for completeness' sake, Fetch's implementation in Firefox works for `file://` scheme.

Besides Chrome, which deliberately implemented its own `fetch` version w/ `file://` scheme removed, dunno about the situation on other browsers yet.

islemaster:
Thanks for clarifying.  I'm pretty sure the issue is not some failure to implement `file://` but a spec'd security limitation of the new API.  In any case, the problem browser isn't Chrome or Firefox but PhantomJS. I presume it's using the [GitHub fetch polyfill](https://github.com/github/fetch) as [provided in p5.js](https://github.com/processing/p5.js/blob/master/src/io/files.js#L15). It gives a status code of `0` and `response.ok` is false when trying to retrieve that file.

GoToLoop:
> I'm pretty sure the issue is not some failure to implement `file://` ...

That's why I've stated it was deliberate.  :P  

> isn't Chrome or Firefox but PhantomJS.

Indeed, Node.js uses Chrome's engine as its default.

RedDuckss:
I've been testing out the changes made in this PR (specifically from here https://github.com/islemaster/p5.play/blob/d6a83d4a017455d557019eaa317083f27e78e4e1/lib/p5.play.js) and I have 2 issues with sprite sheets. Firstly, if 2 sprites are touching, even if you give the proper dimensions for each sprite in the sprite sheet, it seems to add an extra 1 pixel to the sprite from the sprites touching it.

For example, if we have this sprite sheet:
spritesheet1
And we wanted to define just the 32x32 triangle part, the resulting sprite in the browser would actually look like this:
spritesheet2

The only way I have found to fix this is to make it so each sprite in the spritesheet has 1 pixel of border around itself. However that brings me to the second issue.

When a sprite defined from a spritesheet has that pixel of border it adds the transparency from the extra border to sprite, which makes the sprites not seem to connect when loaded.

For example, take this section of a much larger spritesheet I am using:
spritesheet3

When using these sprites, the result is:
spritesheet4
spritesheet5

When what's expected is something more like:
spritesheet6

If you want to verify this for yourselves, the spritesheet I'm using can found here: http://i.imgur.com/AguQgMr.png which is an edited version of http://www.mariouniverse.com/images/sprites/nes/smb/tiles-2.png that I made to split the sections apart.

The code I used for loading the spritesheet sections is:
```javascript
var tile_sheet = loadImage('/assets/textures/tile_sheet.png');

var background_hill_flat_slope_left = new SpriteSheet(tile_sheet, [{'name':'background_hill_flat_slope_left','frame':{'x':264,'y':264,'width':32,'height':32}}]);
background_hill_flat_slope_left = loadAnimation(background_hill_flat_slope_left);

var background_hill_flat_slope_right = new SpriteSheet(tile_sheet, [{'name':'background_hill_flat_slope_right','frame':{'x':330,'y':264,'width':32,'height':32}}]);
background_hill_flat_slope_right = loadAnimation(background_hill_flat_slope_right);

var background_hill_flat_top = new SpriteSheet(tile_sheet, [{'name':'background_hill_flat_top','frame':{'x':297,'y':264,'width':32,'height':32}}]);
background_hill_flat_top = loadAnimation(background_hill_flat_top);

var background_hill_flat_inner = new SpriteSheet(tile_sheet, [{'name':'background_hill_flat_inner','frame':{'x':297,'y':297,'width':32,'height':32}}]);
background_hill_flat_inner = loadAnimation(background_hill_flat_inner);

var background_hill_flat_inner_dots_right = new SpriteSheet(tile_sheet, [{'name':'background_hill_flat_inner_dots_right','frame':{'x':264,'y':297,'width':32,'height':32}}]);
background_hill_flat_inner_dots_right = loadAnimation(background_hill_flat_inner_dots_right);

var background_hill_flat_inner_dots_left = new SpriteSheet(tile_sheet, [{'name':'background_hill_flat_inner_dots_left','frame':{'x':330,'y':297,'width':32,'height':32}}]);
background_hill_flat_inner_dots_left = loadAnimation(background_hill_flat_inner_dots_left);
```

islemaster:
Hi @RedDuckss.  Thank you for testing out the changes in this PR, and sorry it's been sitting around for a while.

I think I understand the issue with sprite sheets you are pointing out, but I'm not sure it's caused by this particular set of changes, or should block us from adding support for p5.js >= 0.5.5.

To check, I threw together a more complete working example [documented in a Gist](https://gist.github.com/islemaster/70b8a69527cec8c1b099a43bf14c1649) using the spritesheet you posted.

I tested this example in two configurations:

|     | p5.js version | p5.play.js version |
| --- | --- | --- |
| Config 1 | 0.5.4 | `master` [7bf47a7](https://github.com/molleindustria/p5.play/commit/7bf47a7e19c68f1c82192f803f74d60e402827ce) |
| Config 2 | 0.5.16 | `p5-g5-0.5.5` [d6a83d4](https://github.com/molleindustria/p5.play/pull/140/commits/d6a83d4a017455d557019eaa317083f27e78e4e1) |

To my eye, both cases render exactly the same:

image

I suspect you have identified a real problem with how we extract tiles from sprite sheets, but it's a problem that exists both before and after this change.  I'd like to consider your issue nonblocking to this pull request, and perhaps open it as a separate issue.

Have I understood you correctly?

ahreyes25:
This still seems to be an issue I'm having with the spritesheets. I've tested with p5 versions: 0.5.5, 0.5.11, & 0.6.0 and still get the Undefined Frame 0 error. I'm not sure if I'm just missing something here? In this build I have the most recent version of p5play and p5js. Here's a link to my code: https://github.com/ahreyes25/p5jsGameEngine/blob/master/src/Player.js
Any help would be appreciated, thanks.

garywu:
This error appears after p5 version 0.5.3 for me. I've created a repo with simplified code to reproduce this error easily. https://github.com/garywu/p5-play-140

quinton-ashley:
As far as I can tell this is no longer an issue. If someone is still having this issue please create a new issue report. Closing the pull request.

PR #144, opened 2017-10-09
Animation - On Complete

hicay:
When animation ends, onComplete method will be fired


hicay:
I deleted forked repo. I can't make new commit, So i created new one
[https://github.com/molleindustria/p5.play/pull/166](url)

PR #150, opened 2017-11-09
Update gh-pages documentation

islemaster:
It's been a while!  Deploying documentation updates, which at least includes the following improvements:

- Update yuidoc-lucid theme, fixing table of contents filtering and example code indentation (#123)
- Fix `Animation.goToFrame` documentation (#111)
- Changes to `Sprite.setCollider` documentation (#112, #118)
- Documentation updates related to `Sprite.friction` change (#120)
- Fix `Animation.nextFrame` and `previousFrame` docs (#133)
- Fix `Animation.frameDelay` documentation (#148)

Once published, this should resolve issues #69, #137, #147, and #152.

You can preview the new docs at http://islemaster.github.io/p5.play/

## Procedure for doing this documentation update:

1. Get a current `gh-pages` branch on my fork (I didn't have one yet)
   ```
   git fetch upstream
   git branch --no-track gh-pages upstream/gh-pages
   ```
2. Make sure my local master branch is current with upstream/master
   ```
   git merge --no-ff upstream/master
   ```
3. Run `update-gh-pages.sh` script while on the master branch.
   ```
   ./update-gh-pages.sh
   ```
   This implicitly switches to the local `gh-pages` branch and stages changes after regenerating documentation.
4. Commit the documentation changes
   ```
   git commit
   ```
5. Push the changes to my fork
   ```
   git push origin gh-pages
   ```
6. Open this pull request from my gh-pages branch into upstream gh-pages branch.

PR #149, opened 2017-11-09
Update GH Pages documentation

islemaster:
It's been a while!  Doing a new documentation deploy to pick up the fix for #147.  Obviously a lot of other stuff gets changed here too.

Procedure for doing this documentation update:

1. Get a current `gh-pages` branch on my fork (I didn't have one yet)
   ```
   git fetch upstream
   git branch --no-track gh-pages upstream/gh-pages
   ```
2. Make sure my local master branch is current with upstream/master
   ```
   git merge --no-ff upstream/master
   ```
3. Run `update-gh-pages.sh` script while on the master branch.
   ```
   ./update-gh-pages.sh
   ```
   This implicitly switches to the local `gh-pages` branch and stages changes after regenerating documentation.
4. Commit the documentation changes
   ```
   git commit
   ```
5. Push the changes to my fork
   ```
   git push origin gh-pages
   ```
6. Open this pull request from my gh-pages branch into upstream gh-pages branch.

PR #151, opened 2017-11-09
Update documentation update procedure

islemaster:
When attempting to update documentation in #149 I discovered the p5.js dependency wasn't being deployed with the docs, due to the change in #139 where we depend on p5.js via npm in order to test against multiple versions.

This makes a slight change to that process and to our update-gh-pages script so that p5.js is deployed with our docs again, so examples work.  The result can be seen in #150.


islemaster:
@caleybrock Would you be willing to review this change sometime today?

caleybrock:
👀

caleybrock:
looks good to me

PR #193, opened 2020-12-16
Draw animation with width and height & separate rotate function

liubq7:
When I used the draw function, I can't set the width and height, so I rewrote the method and separated rotate into another function.


quinton-ashley:
Breaks the intended use of the rotation parameter. You should change the width and height of the image/animation given to the sprite not the sprite itself. Closing request.

PR #215, opened 2022-05-02
Big ES6 update!

quinton-ashley:
This ES6 update to p5.play should be fully backwards compatible! I've tested it with my own projects and it works great. Sprite, Camera, Group (extends Array), CircleCollider, AABB, SpriteAnimation, SpriteSheet, and QuadTree are now ES6 classes.

One of the additional features I added to p5.play is the ability to add animations to groups. Sprites added to a group will be able to use the group's animations when the sprite.changeAnimation function is used. This is highly useful for many use cases.

Another feature is the sprite.autoResetAnimations property.  If false, animations that are stopped before they are completed typically by a call to sprite.changeAnimation, will start at the frame they were stopped at if changed back to. If true, animations will always start playing from frame 0 unless specified by the user in a seperate anim.changeFrame call. I think setting the default to true would be ideal for most use cases but by default this is set to false for backwards compatibility. For example, idle animation of a character lifting its hand to scratch its head (thinking) and putting its hand back down to rest at its side is interrupted in the middle by player movement, player movement ends, idle animation resumes from its starting frame, when the arms are resting at the character's side, not in the middle of the animation which would be undesirable.

### DIFF
--- a/lib/p5.play.js
+++ b/lib/p5.play.js
@@ -4,706 +4,673 @@ by Paolo Pedercini/molleindustria, 2015
 http://molleindustria.org/
 */
 
-(function(root, factory) {
-if (typeof define === 'function' && define.amd)
-define('p5.play', ['p5'], function(p5) { (factory(p5)); });
-else if (typeof exports === 'object')
-factory(require('../p5'));
-else
-factory(root.p5);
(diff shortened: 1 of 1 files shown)


quinton-ashley:
Thanks!

PR #225, opened 2022-08-30
Fix `getSpritesAt` querying for only one fixture.

annalog-codes:
`getSpritesAt` should not stop querying the world when one overlapping fixture is found, but should keep on querying for all overlapping fixtures.


annalog-codes:
This seems to fix the mouse hover problem on stripes with an overlap.
I'm still not sure why the mouse hover was detected when using a group and even then only on certain placements of `.overlap`.

PR #226, opened 2022-08-30
Replace old `s.sensor` with new `s._hasOverlaps`.

annalog-codes:


quinton-ashley:
ah yup forgot to switch those, thank you!

PR #250, opened 2023-07-08
Fps Bounding

ormaq:
Update to make frame rate correct to what it is set to


quinton-ashley:
Thanks for the pull request but in the future you should submit an issue first! In this case `getFPS()` does work as intended.

In gaming fps != frame rate. FPS is a measure of how fast a computer can produce frames. See the docs:

https://p5play.org/docs/global.html#getFPS

Frame rate or display rate is how fast the computer's display can show frames. 60fps is a typical target frame rate.

PR #297, opened 2024-01-26
Update rotate method when handling small angles

sifhg:
Changes line 3312 from:
await pInst.sleep();

to:
return;


quinton-ashley:
Thank you for reporting this issue and for the pull request!

But `rotate` can't just return there because the other code at the end of the function would still need to be run. Check out the fix: https://github.com/quinton-ashley/p5play/blob/main/p5play.js#L3296

PR #314, opened 2024-03-30
Sprite.distance & world.raycast implementation

### LINKED ISSUE #280: `world.rayCast`
Add raycasting.

Override planck's `world.rayCast`

https://piqnt.com/planck.js/RayCast

codingMASTER398:
First time contribution - Some of my formatting may not be up to spec, and the minified version isn't currently included.

This PR aims to fix #280, as well as a pet peeve of mine being the lack of Sprite.distance() function.

## Sprite.distance(otherSprite)
Returns the distance to the other Sprite. How this isn't already implemented is odd to me.

## world.raycast
An implementation atop of the world.rayCast (don't get them confused! I don't know how to override the other one and still use it) that returns the closest sprite from a position and direction of a ray.

Due to limitations I've encountered [in the past](https://github.com/piqnt/planck.js/issues/270), I'm still unable to get Planck to return a list of all rays & find the closest one simultaneously. Even with `Sprite.distance`, I wasn't able to get it working.
Thus, the function currently only returns one "hit" (true/false), "sprite", "point", and "normal".
This can be remedied with the optional `excludeFunction`, detailed in a comment in the PR.

[Here's a sketch demonstrating the functionality](https://editor.p5js.org/codingMASTER398/sketches/x_QnmCD12).

### DIFF
--- a/p5play.js
+++ b/p5play.js
@@ -6797,7 +6797,10 @@ p5.prototype.registerMethod('init', function p5playInit() {
 				result = {
 					hit: true,
 					sprite: body.sprite,
-					point,
+					point: createVector(
+						point.x * plScale,
+						point.y * plScale
+					),
 					normal
 				}
 
--- a/p5play.js
+++ b/p5play.js
@@ -3970,6 +3970,15 @@ p5.prototype.registerMethod('init', function p5playInit() {
 			}
 			this._hasSensors = true;
 		}
+
+		/**
+		 * Returns the distance to another Sprite
+		 * @param {Sprite} otherSprite
+		 * @returns {Number} distance
+		 */
+		distance(otherSprite){
+			return $.dist(this.x, this.y, otherSprite.x, otherSprite.y)
+		}
 	};
 
 	// only used by the p5play-pro Netcode class to convert sprite data to binary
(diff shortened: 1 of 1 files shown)


quinton-ashley:
In p5.js there is already a function for calculating the distance between two points.

`dist` https://p5js.org/reference/#/p5/dist

```js
dist(spriteA.x, spriteA.y, spriteB.x, spriteB.y)
```

There's also `vector.dist`

```js
spriteA.pos.dist(spriteB.pos)
```

But I do like that use of your function would be far simpler. There's already `sprite.angleTo` so perhaps for consistency it should be called `sprite.distanceTo`

```js
spriteA.distanceTo(spriteB)
```

It also might be nice to make p5's `dist` function accept sprites, mouse, and any object with x,y properties.

```js
dist(spriteA, mouse)
```

codingMASTER398:
`sprite.distanceTo()` is a good idea- I'll add another commit changing the name of `sprite.distance()`

quinton-ashley:
Also I forgot to say since p5play's `World` class extends `planck.World` you can override the `world.rayCast` method and then still use `planck.World`'s `rayCast` inside it by using `super` which you can see an example of in my `world.step` function. I'll make the edits needed to do that.

Merging your pull request! Thank you!

quinton-ashley:
Also one important feature of the raycast you didn't include is being able to end it. For example, if we don't want a projectile to go through a wall, then we don't need to check behind the wall for better performance. I will implement this.

quinton-ashley:
I'm going to rename the `excludeFunction` param to `filter` and returning true will include the sprite in a list of results, like [Array.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).

quinton-ashley:
I also want `world.rayCast` to return a list of all the sprites the ray intersects with. I will try implementing this.

quinton-ashley:
I like that you used radius and direction for this function instead of a second point. Great for p5play ease of use!

quinton-ashley:
@codingMASTER398 I added some additional info to your planck issue report: https://github.com/piqnt/planck.js/issues/270

quinton-ashley:
I'm going to switch the input param order to match Unity's as well.

https://docs.unity3d.com/ScriptReference/Physics.Raycast.html
https://docs.unity3d.com/ScriptReference/Physics.RaycastAll.html

PR #328, opened 2024-07-25
Update p5play.d.ts

sifhg:
The method `applyForce` can take only one argument. More than that is optional.


quinton-ashley:
true! But the p5play.d.ts file is generated by the jsdoc comments in p5play.js. I'll make the change there as well.

quinton-ashley:
Thanks for pointing out the issue! Fixed in v3.22.12

PR #330, opened 2024-09-10
add `repelFrom()` function to `Group`

Dukemz:
This is my first time contributing, please let me know if I messed something up.
Since there's an `attractTo()` for groups I can't think of any reason why this shouldn't exist.


quinton-ashley:
Good addition, will pull this soon!

PR #340, opened 2025-02-14
Hotfix for image loading callbacks in Q5

codingMASTER398:
See https://discord.com/channels/1003099076960723016/1007771376263905391/1339887894210281502
An edge-case with Q5 & image loading callbacks that I probably introduced in my other PR, lol. This should fix it.


quinton-ashley:
Good catch, I think just checking the image width and height is fine though.

